PR #22, opened 2022-07-21
inital commit

austin3dickey:


jonkeane (commented):
Could we also add a license + copyright headers to our files (with Voltron Data as the copyright holder)?

austin3dickey:
> Could we also add a license + copyright headers to our files (with Voltron Data as the copyright holder)?

You had already added the LICENSE. 🙂 I just added a copyright header to all files except

- LICENSE (obviously)
- poetry.lock because that's autogenerated
- README.rst (comments are hard in RST and it seems unnecessary)
- repo.json (comments are hard in JSON)

jonkeane (approved):
Thanks! 

> README.rst (comments are hard in RST and it seems unnecessary)

Any reason not to use markdown for this instead of rst?

austin3dickey:
> Any reason not to use markdown for this instead of rst?

RST integrates a little better with Sphinx if we want to set that up in the future, and I think it's a little more feature-rich than Markdown. But I know a lot more people are familiar with writing Markdown. (Comment blocks are equally hacky in both of them.) 

I just continued with RST because I didn't want to refactor what @joosthooz had written originally.

PR #26, opened 2022-07-22
Added CI step to check files for copyright headers

### LINKED ISSUE #23: Add a step to the build that checks copyright headers
So we don't forget, let's add a step to check all applicable files for the correct copyright/license header.

austin3dickey:
This PR adds a check that our copyright information looks correct. Fixes #23 . cc @joosthooz since I can't add you as a reviewer yet.


ci/check_copyright.py
```diff
@@ -0,0 +1,72 @@
...
+    if current_year == 2022:
+        header = HEADER.format(years="2022")
+    else:
+        header = HEADER.format(years=f"2022-{current_year}")
+
+    violations = []
+    for file in files:
+        if str(file) not in EXCLUSIONS:
+            with open(file, "r") as f:
+                if not f.read().startswith(header):
```
austin3dickey, line 64: This is currently pretty strict—that files start with this exact block of text—but it works for all current files. We can revisit it to be more complex/flexible if we need e.g. a different comment style.

jonkeane (commented):
Thanks for this! ❤️ doing this now instead of dealing with forgetting later. Looking just at number of lines, copying https://github.dev/apache/arrow/blob/master/.github/workflows/dev_pr.yml would be the same number and presumably handles more different kinds of files? Would using the apache-rat be feasible here out of the box?

austin3dickey:
>copying https://github.dev/apache/arrow/blob/master/.github/workflows/dev_pr.yml would be the same number and presumably handles more different kinds of files

I don't see anything in the file you linked about license checking or Rat; am I missing something or did you mean to link a different file?

It looks like `arrow` uses all [this](https://github.com/apache/arrow/blob/master/dev/archery/archery/utils/rat.py) code, along with [this](https://github.com/apache/arrow/blob/master/dev/archery/archery/utils/lint.py#L323-L344) code and [this](https://github.com/apache/arrow/blob/master/.github/workflows/dev.yml#L52-L54) code to run Rat. But I think we could probably simplify a lot of that. Let me take a crack at it.

austin3dickey:
I overhauled the script to parse a Rat report. Check it out!

jonkeane (approved):
Looks good! And oops, sorry I _meant_ to send a link to exactly that archery code!

PR #27, opened 2022-07-22
Added a pre-commit with black, flake8, and isort

### LINKED ISSUE #5: Add a pre-commit hook for developers
Similar to other projects in this organization, we should add a pre-commit hook for linting with black, isort, etc. And fail the build when it is not satisfied.

austin3dickey:
This PR adds black, flake8, and isort as linters, with an associated pre-commit configuration and CI step. It also applies isort and fixes some flake8 problems.

Fixes #5 . cc @joosthooz since I can't add you as a reviewer yet.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Create poetry environment
         run: poetry install --no-root
       - name: Lint
-        run: pre-commit run --all-files
+        run: poetry run pre-commit run --all-files
   unit-tests:
     strategy:
       matrix:
@@ -52,7 +52,7 @@ jobs:
       - name: Create poetry environment
         run: poetry install
       - name: Run tests
-        run: pytest
+        run: poetry run pytest
   build-wheel:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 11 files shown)


.github/workflows/actions.yml
```diff
@@ -21,6 +21,19 @@ on:
...
+      - name: Check out repo
+        uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+      - name: Install Poetry
+        run: pip install poetry
+      - name: Create poetry environment
+        run: poetry install --no-root
+      - name: Lint
+        run: poetry run pre-commit run --all-files
```
jonkeane, line 36: We don't have to do this now, but it's possible to have a command that runs these and pushes the results back? Similar to [Apache Arrow's autotune](https://github.com/apache/arrow/blob/4e37149615da8f55b7e72f469248834d98502c7b/.github/workflows/comment_bot.yml#L54-L151). Think we could add that (or make an issue for adding that?)

austin3dickey: We could do `black` and `isort` automatically, but automatically solving problems exposed by `flake8` ranges from tricky to impossible. I'm hoping that `pre-commit` will prevent most cases of pushing unlinted code to the repo. But yep, I just made https://github.com/conbench/datalogistik/issues/30 because it sounds like fun to try.

jonkeane (approved):

Issue #19, opened 2022-07-20
Create a cookbook for integrating datalogistik into a benchmark runner


austin3dickey:
Have examples people can copy/paste to run datalogistik from their language of choice and reference its files

jonkeane:
I've bumped this from the mvp. I love the idea + want to have a cookbook, but let's get datalogistik into a few places and make sure we've got the API right before we invest in this IMO

PR #28, opened 2022-07-25
Switched all structlogs/prints to stdlib logger

### LINKED ISSUE #7: Fix logger
Let's get rid of `structlog` and use the built-in `logging` module

austin3dickey:
This PR switches all print statements to the standard library logger, and removes the structlog dependency. It standardizes how we handle logging in the package. Fixes #7

### DIFF
--- a/datalogistik/cli.py
+++ b/datalogistik/cli.py
@@ -15,11 +15,11 @@
 import argparse
 import json
 import os
-import sys
 
 import urllib3
 
-from . import config, tpc_info, util
+from . import config, tpc_info
+from .log import log
 
 
 def parse_args():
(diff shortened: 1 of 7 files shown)


jonkeane (approved):
This is looking good — should we document somewhere where this output goes | how to access it? (or is it still to stdout?)

austin3dickey:
> where this output goes | how to access it? (or is it still to stdout?)
Yep, it's still just to stdout. In the future if we wanted it's easy to write to a file instead or in addition. (Maybe we'll think about that when doing #19, because it'd be interesting to just print the cache location to stdout and put logs in a file.)

jonkeane:
Sounds good. I'm fine to merge this now (well, after the conflicts are resolved)

PR #29, opened 2022-07-25
Added a builder for `dsdgen`

### LINKED ISSUE #3: Provide a builder for `dsdgen`
Similar to our builder for `dbgen`, we should add a class that clones/builds the TPC-DS tool if users don't have it.

austin3dickey:
This PR adds support for building the `dsdgen` TPC-DS generator from scratch if a user doesn't have it. It will clone the tool from https://github.com/gregrahn/tpcds-kit/.

The code for cloning/making the tool is really similar to the TPC-H code, so in this PR I shoved a lot of that code back into the parent `_TPCBuilder` class. This makes the diff kinda hard to read; I'd recommend just reading the new `tpc_builders.py` file.

The PR also moves the TPC integration tests outside of the "test different methods of installation" checks so that those run faster and we're using less compute during the build.

Fixes #3 .

### DIFF
--- a/datalogistik/tpc_builders.py
+++ b/datalogistik/tpc_builders.py
@@ -246,7 +246,9 @@ class DSDGen(_TPCBuilder):
         log.info("Upgrading the solution file; this could take a few minutes...")
         solution_file = self.repo_build_path / "dbgen2.sln"
 
-        # Remove unnecessary projects from this file
+        # See https://github.com/gregrahn/tpcds-kit/blob/5a3a817/tools/dbgen2.sln
+        # Some of these projects (dsqgen, checksum, and grammar) don't upgrade nicely.
+        # We don't need them, so delete all references to them from this file.
         with open(solution_file, "r") as f:
             new_solution = "".join(
                 line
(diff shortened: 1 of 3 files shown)


austin3dickey:
still a draft, no review required yet

datalogistik/tpc_builders.py
```diff
@@ -215,16 +213,55 @@ class DSDGen(_TPCBuilder):
...
+        log.info("Upgrading the solution file; this could take a few minutes...")
+        solution_file = self.repo_build_path / "dbgen2.sln"
+
+        # Remove unnecessary projects from this file
+        with open(solution_file, "r") as f:
+            new_solution = "".join(
+                line
+                for line in f
+                if not any(
+                    uuid in line for uuid in ("59EBAD48", "3EA62CB9", "6540812A")
```
joosthooz, line 255: What are these numbers?

joosthooz (approved):
Nice, thanks!

datalogistik/tpc_builders.py
austin3dickey: Good question. [This](https://github.com/gregrahn/tpcds-kit/blob/5a3a8/tools/dbgen2.sln) is the solution file that builds the tool on Windows. It's actually supposed to build 6 different tools, including `dsdgen`. But I found that it takes a long time and sometimes crashes our CI to build all 6 tools. I think something is wrong with one or two of them. 

So this code deletes all references to `dsqgen`, `checksum`, and `grammar` in this file so that it only builds `dsdgen` (and its direct dependencies `distcomp` and `mkheader`). This makes everything work again.

I'll add a comment to make this clearer.

PR #31, opened 2022-08-04
Added cache bypass option

joosthooz:
PR to address #10 
Subtasks:
- [x] An option to bypass the caching mechanism for intermediate formats (preferably keeping disk usage to a minimum, so immediately removing files once not needed anymore) -> [d8216fe](https://github.com/conbench/datalogistik/pull/31/commits/d8216fe1c039f0822ddc802fa347ee9d69440f0f)
- [x] An option to remove cache entries by literal specifying the subpath to remove

Issue #25, opened 2022-07-22
Add Sphinx documentation


austin3dickey:
Set up Sphinx documentation for the package, and optionally host it on Github Pages.

PR #34, opened 2022-08-05
validate file integrity

### LINKED ISSUE #1: Validate file integrity
Historically {ursacomputing/benchmarks} has validated file integrity by checking file size equals an expected size. Because a lot of factors can impact file size, this became a pain, but the idea of validating that files used for benchmarks are consistent is still worthwhile, if we have a robust way to test it. This PR removed the validation, showing exactly what was done: https://github.com/ursacomputing/benchmarks/pull/117#discussion_r923849121

### LINKED ISSUE #35: Validate file integrity from repo file
Continuation of #1 
In addition to validating whether the contents of a dataset were changed since they were downloaded (by comparing against the checksum generated by datalogistik itself), we should add the option to add a checksum to dataset properties in a datasources repo file. That way we can validate if the datasource was properly downloaded in the first place.

joosthooz:
resolves #1 
resolves #35 

Added a `validate` command that can:
- [x] => no longer doing this: _Verify the integrity of a dataset given by path_
- [x] Verify the integrity of all the entries in the cache

In addition; now checking checksums after downloading a file, if there is a checksum in the repo file


austin3dickey:
I'll review this once we merge #31 ; the diff is a little confusing otherwise

.github/workflows/actions.yml
```diff
@@ -119,8 +119,40 @@ jobs:
           name: poetry-build-artifacts
       - name: Install package
         run: ${{ matrix.installer }} install '${{ matrix.installee }}'
-      - name: Test generating and converting a dataset
+      - name: Test Downloading a dataset
         run: datalogistik generate -d type_floats -f parquet
+  test-download-convert-p2c-without-repo:
```
austin3dickey, line 124: This new section feels a little redundant to me. I should have put more comments in this file, but what I was thinking for the CI strategy was:
- `lint`, `check-copyright`: self-explanatory
- `unit-tests`: test as much behavior of this package as possible, without calling the CLI
- `build-wheel -> test-convert-without-repo`: a simple integration test of the CLI without cloning the repo, in order to test that that workflow works
- `test-tpc-integration`: an integration test of TPC-H and TPC-DS because they're complicated and need additional setup than we could provide in `unit-tests`

If you want to test that the conversion works, I'd say your unit tests are good without this new section in this file. But if you wanted to be extra sure, you could also change line 123 to "csv" instead of "parquet". I don't think we need another whole step though.

datalogistik/cli.py
```diff
@@ -31,6 +31,7 @@ def parse_args():
     sub_parsers = parser.add_subparsers(dest="command")
     cache_parser = sub_parsers.add_parser("cache")
     gen_parser = sub_parsers.add_parser("generate")
+    valid_parser = sub_parsers.add_parser("validate")
```
austin3dickey, line 34: Could you please add documentation to the README? 🙂

datalogistik/util.py
```diff
@@ -83,6 +84,73 @@ def add_file_listing(metadata, path):
...
+        log.error(msg)
+        raise RuntimeError(msg)
+    if not path.is_dir():
+        msg = f"Path '{path}' is not a directory"
+        log.error(msg)
+        raise RuntimeError(msg)
+    metadata_file = pathlib.Path(path, config.metadata_filename)
+    dataset_found = False
+    if metadata_file.exists():
+        orig_file_listing = json.load(open(metadata_file)).get("files")
```
austin3dickey, line 101: Could you please switch to the contextmanager `with open()` pattern here?

tests/test_datalogistik.py
```diff
@@ -101,8 +102,37 @@ def test_convert_dataset_csv_to_parquet():
...
     print(converted_table.schema)
     assert converted_table == orig_table
-    (test_csv_file_path).unlink()
-    (test_parquet_file_path / "part-0.parquet").unlink()
-    (path / config.metadata_filename).unlink()
-    (converted_path / config.metadata_filename).unlink()
     shutil.rmtree("./test_csv", ignore_errors=True)
+
+
+def test_convert_dataset_parquet_to_csv():
```
austin3dickey, line 108: Thanks for this test! Could you add a test that the validation functionality works?

austin3dickey (changes requested):

datalogistik/util.py
```diff
@@ -214,7 +291,7 @@ def get_dataset(input_file, dataset_info, table_name=None):
     if dataset_info["format"] == "csv":
         # defaults
         po = csv.ParseOptions()
-        ro = csv.ReadOptions()
+        ro = csv.ReadOptions()  # autogenerate_column_names=True)
         co = csv.ConvertOptions()
```
alistaire47, line 295: If the source of truth is in csv which then gets converted on to parquet and whatever, these options can have a non-trivial impact on what gets written back out to other formats. [In benchmarks](https://github.com/voltrondata-labs/benchmarks/blob/main/benchmarks/_sources.py#L329-L354) I did a bit to lock it down, though I'd still hesitate to call it robust.

And for writing parquet [you may need to set `coerce_timestamps`](https://github.com/voltrondata-labs/benchmarks/blob/main/benchmarks/_sources.py#L432) for consistent timestamp precision depending on what parquet version you're using.

repo.json
```diff
@@ -6,7 +6,7 @@
     "dim" : [22180168, 31],
     "format" : "csv",
     "file-compression" : "gz",
-    "schema" : {
+    "tables": [{"table" : "fanniemae_2016Q4.csv", "schema" : {
```
alistaire47, line 9: [Here's a better schema for fanniemae](https://github.com/voltrondata-labs/benchmarks/blob/main/benchmarks/_sources.py#L54-L98) whenever you want it

alistaire47 (approved):
Looks good! A few comments to highlight some landmines in data validation that I've discovered the hard way to be aware of as you move forward

datalogistik/util.py
```diff
@@ -83,6 +84,73 @@ def add_file_listing(metadata, path):
     metadata["files"] = file_list
 
 
+# Validate that the integrity of the files in the dataset at given path is ok, using the metadata file
+def validate(path):
```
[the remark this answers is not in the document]
joosthooz: 1) Thanks for spotting this! I will take care of this.
2) Yes, also individual files will be placed in a directory, also because there will always be a metadata file added. The presence of that metadata file (and the "files" entry in it) will be the factor that determines whether the dataset seems valid (if the program finds a directory without it, it will delete the directory).
3) I don't check them because if the file size is different (so the files are not identical), you can bet your butt that the checksums are also different. But I can add a size check before the checksum check

datalogistik/util.py
joosthooz: Thanks, I'm actually struggling with these options because the schema detection in pyarrow.dataset is not good enough for the test datasets (fanniemae and nyctaxi). Passing the names would already help a bit but will not solve problems with the types of the schema: dataset detects the schema from the first x bytes of data (I think 1MB). So if there is a different value somewhere later on, it results in an error.
In the end a real solution would be to add the schema into the repo file and convert it to a pyarrow schema and pass it to the call creating the dataset. That's for #36 .

repo.json
joosthooz: This is awesome, I'm going to add it right away

joosthooz: [d3e9d58](https://github.com/conbench/datalogistik/pull/34/commits/d3e9d5827d019a733e6efb5b262e0012795182a7)

tests/test_datalogistik.py
joosthooz: Done in [88c8faf](https://github.com/conbench/datalogistik/pull/34/commits/88c8fafcaa550b85c462d0f61db1ac46dca3212f)

datalogistik/util.py
joosthooz: Done, also for 1 spot in the unit tests

datalogistik/cli.py
joosthooz: Yes! I Forgot (again), added

.github/workflows/actions.yml
joosthooz: I understand, and removed the new test! Also changed output type to csv, to include a conversion. Maybe I'll add some more commands in this test, just so it hits most of the code

datalogistik/util.py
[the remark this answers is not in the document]
joosthooz: The deduplication is done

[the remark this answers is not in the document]
joosthooz: Ok I looked at the integrity check code again, and it does check both file size and checksum, because it just checks the whole file listing array item (containing file path, size, and md5). This check is performed 1) after downloading and decompressing a dataset, but only in case there is a file listing for it in the repo file, and 2) at user request (either for all the entries in the cache, or for a given path)

datalogistik/util.py
alistaire47: Yep, it's almost necessary to specify a schema when the original form is CSV/delimited; I ran into lots of issues with coerced nulls and null-type columns that caused downstream issues in addition to the obvious type differences

datalogistik/util.py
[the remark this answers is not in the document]
alistaire47: Cool, size/checksum checks on the downloads should be pretty safe. Sizes get dicier for converted versions (e.g. a parquet/snappy/1e6 chunk size version of csv/gzip original) because they depend on all the reading and writing options, schema, and behavior, so any changes, deliberate or in defaults or behavior, cause failures. Less of an issue if deps are pinned, but I think the default parquet version is in the midst of changing, so choice of pin may be important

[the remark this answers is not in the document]
joosthooz: Yes, checking generated/converted datasets will be very difficult. We would need to extract only the pure contents and have a checksum for that in the repo file too. That's probably a bit outside the scope of this tool.

datalogistik/util.py
joosthooz: I'm looking into coerce_timestamps, but `pyarrow.dataset` does not have en equivalent option.

datalogistik/util.py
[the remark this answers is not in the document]
jonkeane: Yeah, agreed that the derived datasets is probably more complicated than it's worth (certainly for now possible forever). It also should be _relatively_ quick to re-generate the derived datasets anyway, such that if we're having issues we can prune the cache and then everything should be fine.

jonkeane:
Thanks for this! I haven't dug in too deeply, but I do wonder if we even need the ad-hoc verification command to be surfaced at all. In normal operations I would expect this to run after a download (which it is already), and that's probably about it. 

It's nice to have the ability to do this adhoc, but I wonder if it's adding more user interface than we really need to at this point?

joosthooz:
You may be right. I added it because some people might want to do read and writes to a dataset, but maybe that doesn't make sense and people should just re-instantiate in that case. If the dataset is in the cache that would be comparable to verifying. I'm ok with removing the command. Maybe we can just merge cache cleaning and verification into 1 option that does both?

joosthooz:
There is now no longer a validate subcommand, instead there is a cache --validate command that verifies file integrity of all the cache entries. Invalid entries are pruned.

README.rst
```diff
@@ -9,7 +9,7 @@ generate some datasets such as TPC-H by calling their (external) generator progr
 
 Usage::
 
-    datalogistik <generate|cache>
+    datalogistik <generate|cache|validate>
```
jonkeane, line 12:
```suggestion
    datalogistik <generate|cache>
```

If we only have an argument now we should do this, yeah?

datalogistik/cli.py
```diff
@@ -44,6 +44,11 @@ def parse_args():
         action="store_true",
         help="Remove any incomplete/left-over directories from the cache",
     )
+    cache_group.add_argument(
+        "--validate",
+        action="store_true",
+        help="Validate all entries in the cache for file integrity",
+    )
```
jonkeane, line 51: If this auto-prunes, maybe we should call this `--prune-invalid` or something? That way it's clear what it's going to do if there's something that is invalid?

jonkeane (commented):
A few small comments and one big one: 

I thought the last time I looked at this, we had code that would validate the single source of data after it had been downloaded. But I'm struggling to find that again. Did we remove that? IMHO, validating after downloading is probably the only thing we want to expose right now (and do it unconditionally). Most datasets should be relatively easy to download such that if someone is concerned about their cache being corrupt, we should make it easy for them to delete everything for a datasource and start from scratch (downloading it).

Thoughts?

README.rst
austin3dickey: Agreed; looks like you'll need to update the whole README again

austin3dickey (changes requested):
README

joosthooz:
Sorry for forgetting the README again.
@jonkeane, validate is always performed if the repo file contains a file listing (with md5sums) for the downloaded dataset. But the location of that check was a bit strange (in `write_metadata()`), so I moved it to `download_dataset()` now.
I split the --validate option into --validate, that only reports if datasets are invalid, and --prune-invalid, that also deletes them.

austin3dickey:
>Sorry for forgetting the README again

haha no problem; once we do https://github.com/conbench/datalogistik/issues/25 we could have the docs automatically just print out the help page so we don't have to remember to update the README over and over

austin3dickey (approved):
I think this makes sense; I'm glad the validation is done upon download. I'm not positive anyone will ever need to use `--prune-invalid` or `--validate`, but if we don't want users to touch the cache directly then it makes sense to provide an approved way to do those things.

Issue #36, opened 2022-08-10
Add functionality to specify schema


joosthooz:
Add code to read the schema from the datasources repo file, or possibly even allow users to specify a certain schema.

alistaire47:
Is there a good/preferred way of saving schemas? It'd be really nice if we could take them to/from json, but thus far I haven't seen a reasonably robust way of doing so

alistaire47:
We might need to go a little deeper than that for fields with options, e.g. unit/timezone for timestamps and precision/scale for decimal. Maybe just optional kwargs that can be passed through like

```json
{
  "id": {"type_name": "string"},
  "click_ts": {"type_name": "timestamp", "type_args": {"unit": "ms"}},
  "some_value": {"type_name": "decimal128", "type_args": {"precision": 7, "index": 3}},
  ...
}
```

Or flat or otherwise—I don't really care about arrangement, just that it's flexible enough when we need it

joosthooz:
This is a really good point, I'm copying this into the PR so I can easily find it, thanks!

joosthooz:
PR: #41

jonkeane:
I haven't dug too deeply into the PR yet, but https://github.com/apache/arrow/issues/13803 has some discussion about this from the Arrow perspective + some links to how this (apparently?) works in Rust. We certainly would benefit from following that as closely as possible (or possibly upstreaming the work | reigniting the conversation on the ML, etc if we think it's important to push into Arrow itself).

https://github.com/apache/arrow/pull/7110 also had a PR to do some of this as well (though it was closed due to a lack of pushing forward + enthusiasm).

joosthooz:
Thanks @jonkeane , it seems there are more people trying to build more or less this same thing. Maybe I can just translate this piece of code to python https://github.com/apache/arrow-rs/blob/27f4762c8794ef1c5d042933562185980eb85ae5/arrow/src/datatypes/datatype.rs#L785 . If it ends up becoming part of Arrow itself, we can always convert to just calling the API function.

jonkeane:
> If it ends up becoming part of Arrow itself, we can always convert to just calling the API function.

💯 absolutely. We also do have a slightly more constrained scope than arrow, so it's possible the rabbit hole of "what to do about embedded types" won't be so bad for us?

joosthooz:
[...]

PR #37, opened 2022-08-10
Add end2end test for downloading & csv -> parquet conversion

joosthooz:
This PR will need to wait until we have functionality to specify a schema in the repo file (#36). Otherwise, pyarrow.dataset will incorrectly auto-detect the schema.


joosthooz:
Ok, after reading your comment I see this is a bit much. Maybe we can just expand the existing e2e test with a couple of commands and fail if any of them fail.
I'm closing this PR

PR #38, opened 2022-08-10
Added homepage to datasource properties

joosthooz:


joosthooz:
Instead of "original url" I choose to just use "homepage", because is it more general and probably has more information about the origins of the dataset. Also often you need to create an account or request access before you get a download link.

austin3dickey (approved):

PR #39, opened 2022-08-10
13 solidify cache hierarchy level names

### LINKED ISSUE #13: Solidify cache hierarchy level names
- [x] Dataset API creates a directory with given name, containing part-### files. When downloading or generating, the files are created directly. This should be unified (probably: no partitioning -> plain files, otherwise directory with part-### files).
- [x] Come up with better names for the cache hierarchy level names (e.g. partitioning-by-rows-### and scale-factor-###)

joosthooz:
Resolves #13 
- Renames unpartitioned dataset files back to their original filenames instead of creating a directory with a `part-0.format` file
- Added prefixes to the cache directory structure

### DIFF
--- a/datalogistik/util.py
+++ b/datalogistik/util.py
@@ -285,7 +285,7 @@ def convert_dataset(
         cached_dataset_path, config.metadata_filename
     )
     if not cached_dataset_metadata_file.exists():
-        msg = "Could not find source dataset"
+        msg = f"Could not find source dataset at {str(cached_dataset_metadata_file)}"
         log.error(msg)
         raise ValueError(msg)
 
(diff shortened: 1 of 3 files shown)


datalogistik/util.py
```diff
@@ -323,6 +329,18 @@ def convert_dataset(
                 max_rows_per_group=new_nrows if new_nrows != 0 else None,
                 file_visitor=file_visitor if config.debug else None,
             )
+            if new_nrows == 0:
+                # Convert from name.format/part-0.format to simply a file name.format
+                # To stay consistent with downloaded/generated datasets (without partitioning)
+                tmp_dir_name = pathlib.Path(
```
austin3dickey, line 335: You don't have to change any of these, but just FYI: doing `pathlib.Path(my_path, "something_else")` is the same as `my_path / "something_else"` if `my_path` is a `Path` object. It's OS-agnostic too!

austin3dickey (approved):
looks good!

datalogistik/util.py
joosthooz: Yes I've seen that funny `/` overload, but I often feel unsure if `my_path` is already a pathlib so I just use this to be sure :sweat_smile: 
When we add type hints to the functions this would be easier.

PR #41, opened 2022-08-11
Add functionality to specify schema

joosthooz:
Instead of relying on pyarrow.dataset, we would like to be able to specify a schema in the repo file. This will help solve type problems in large files, because pyarrow.dataset only inspects the first 1MB of data to infer types of columns.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -211,12 +211,13 @@ In addition, entries can have the following optional properties:
     - type_name: Name of an Arrow datatype
     - arguments: either a dictionary of argument_name:value items, a list of values,
     or a single value.
-    Example::
+    Example:
+    .. code::
 
     {
-    "a": "string",
-    "b": {"type_name": "timestamp", "arguments": {"unit": "ms"}},
-    "c": {"type_name": "decimal", "arguments": [7, 3]}
+        "a": "string",
+        "b": {"type_name": "timestamp", "arguments": {"unit": "ms"}},
+        "c": {"type_name": "decimal", "arguments": [7, 3]}
     }
 
 ``header-line``
--- a/README.rst
+++ b/README.rst
@@ -213,9 +213,11 @@ In addition, entries can have the following optional properties:
     or a single value.
     Example::
 
-    {"a": "string",
+    {
+    "a": "string",
     "b": {"type_name": "timestamp", "arguments": {"unit": "ms"}},
-    "c": {"type_name": "decimal", "arguments": [7, 3]}}
+    "c": {"type_name": "decimal", "arguments": [7, 3]}
+    }
 
 ``header-line``
     Boolean denoting whether the first line of a CSV file contains the column names (default: false)
--- a/README.rst
+++ b/README.rst
@@ -212,7 +212,7 @@ In addition, entries can have the following optional properties:
     - arguments: either a dictionary of argument_name:value items, a list of values,
     or a single value.
     Example:
-    .. code::
+.. code::
 
     {
         "a": "string",
--- a/README.rst
+++ b/README.rst
@@ -204,7 +204,10 @@ In addition, entries can have the following optional properties:
     but that is not what is meant here.
 
 ``schema``
-    The schema of the tabular data in the file. This entry is currently ignored.
+    The schema of the tabular data in the file.
+
+``header-line``
+    Boolean denoting whether the first line of a CSV file contains the column names (default: false)
 
 Dataset output
 --------------
--- a/README.rst
+++ b/README.rst
@@ -205,6 +205,19 @@ In addition, entries can have the following optional properties:
 
 ``schema``
     The schema of the tabular data in the file.
+    The structure of a schema is a JSON string with key:value pairs for each column.
+    The key is the column name, and the value is either the name of an Arrow datatype
+    without any parameters, or a dictionary with the following properties:
+    - type_name: Name of an Arrow datatype
+    - arguments: either a dictionary of argument_name:value items, a list of values,
+    or a single value.
+    Example::
+
+    {
+      "a": "string",
+      "b": {"type_name": "timestamp", "arguments": {"unit": "ms"}},
+      "c": {"type_name": "decimal", "arguments": [7, 3]}
+    }
 
 ``header-line``
     Boolean denoting whether the first line of a CSV file contains the column names (default: false)
--- a/README.rst
+++ b/README.rst
@@ -213,11 +213,9 @@ In addition, entries can have the following optional properties:
     or a single value.
     Example::
 
-    {
-      "a": "string",
-      "b": {"type_name": "timestamp", "arguments": {"unit": "ms"}},
-      "c": {"type_name": "decimal", "arguments": [7, 3]}
-    }
+    {"a": "string",
+    "b": {"type_name": "timestamp", "arguments": {"unit": "ms"}},
+    "c": {"type_name": "decimal", "arguments": [7, 3]}}
 
 ``header-line``
     Boolean denoting whether the first line of a CSV file contains the column names (default: false)
--- a/datalogistik/util.py
+++ b/datalogistik/util.py
@@ -399,9 +399,7 @@ def get_dataset(input_file, dataset_info, table_name=None):
         po = csv.ParseOptions()
         ro = csv.ReadOptions()
         co = csv.ConvertOptions()
-        # TODO: Should we autogenerate column names by default?
-        # Or add a property in the metadata about it?
-        # or allow a fall-back to read_csv in case schema detection fails?
+        # TODO: Should we fall-back to read_csv in case schema detection fails?
 
         if "delim" in dataset_info:
             po = csv.ParseOptions(delimiter=dataset_info["delim"])
@@ -436,7 +434,6 @@ def get_dataset(input_file, dataset_info, table_name=None):
                 has_header_line = dataset_info.get("header-line", False)
                 autogen_column_names = not has_header_line
 
-            log.debug(f"autogen_column_names = {autogen_column_names}")
             ro = csv.ReadOptions(
                 column_names=column_names,
                 autogenerate_column_names=autogen_column_names
--- a/datalogistik/util.py
+++ b/datalogistik/util.py
@@ -389,9 +389,7 @@ def get_dataset(input_file, dataset_info, table_name=None):
                 for table_entry in dataset_info.get("tables"):
                     if table_name is None or table_entry["table"] == table_name:
                         schema = get_arrow_schema(table_entry["schema"])
-                        column_names = list(
-                            table_entry["schema"].keys()
-                        )  # TODO: does this really always preserve ordering?
+                        column_names = list(table_entry["schema"].keys())
                         break
                 ro = csv.ReadOptions(column_names=column_names)
 
--- a/datalogistik/util.py
+++ b/datalogistik/util.py
@@ -356,8 +356,11 @@ def get_dataset(input_file, dataset_info, table_name=None):
     if format == "csv":
         # defaults
         po = csv.ParseOptions()
-        ro = csv.ReadOptions()  # autogenerate_column_names=True)
+        ro = csv.ReadOptions()
         co = csv.ConvertOptions()
+        # TODO: Should we autogenerate column names by default?
+        # Or add a property in the metadata about it?
+        # or allow a fall-back to read_csv in case schema detection fails?
 
         if "delim" in dataset_info:
             po = csv.ParseOptions(delimiter=dataset_info["delim"])
@@ -491,13 +494,6 @@ def convert_dataset(
                     "schema": schema_to_dict(dataset.schema),
                 }
             )
-
-            # TODO: The dataset API does a poor job at detecting the schema.
-            # Would be nice to be able to fall back to read/write_csv etc.
-            # Another option is to store the schema as metadata in the repo and pass it
-            # to dataset.
-            # It would also be nice to detect/provide option whether the first line
-            # contains column names.
 
         conv_time = time.perf_counter() - conv_start
         log.info("Finished conversion.")
--- a/datalogistik/util.py
+++ b/datalogistik/util.py
@@ -488,7 +488,10 @@ def convert_dataset(
                 if parquet_compression is None:
                     parquet_compression = "snappy"  # Use snappy by default
                 write_options = dataset_write_format.make_write_options(
-                    compression=parquet_compression
+                    compression=parquet_compression,
+                    use_deprecated_int96_timestamps=False,
+                    coerce_timestamps="us",
+                    allow_truncated_timestamps=True,
                 )
             if new_format == "csv":
                 dataset_write_format = ds.CsvFileFormat()
--- a/datalogistik/util.py
+++ b/datalogistik/util.py
@@ -349,7 +349,6 @@ def arrow_type_from_json(input_type):
     if isinstance(args, Mapping):
         return arrow_type_function_lookup(type_name)(**args)
     elif isinstance(args, list):
-        log.debug(f"args {args}")
         return arrow_type_function_lookup(type_name)(*args)
     else:  # args is probably a single value
         return arrow_type_function_lookup(type_name)(args)
--- a/datalogistik/util.py
+++ b/datalogistik/util.py
@@ -375,6 +375,17 @@ def get_arrow_schema(input_schema):
     return output_schema
 
 
+# Lookup the user-specified schema or return None is none was found
+def get_table_schema_from_metadata(dataset_info, table_name):
+    if dataset_info.get("tables"):
+        for table_entry in dataset_info.get("tables"):
+            if table_name is None or table_entry["table"] == table_name:
+                if table_entry.get("schema"):
+                    return table_entry["schema"]
+                break # there should be only 1
+    return None
+
+
 # Create Arrow Dataset for a given input file
 def get_dataset(input_file, dataset_info, table_name=None):
     # Defaults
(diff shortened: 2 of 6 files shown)


joosthooz:
>We might need to go a little deeper than that for fields with options, e.g. unit/timezone for timestamps and precision/scale for decimal. Maybe just optional kwargs that can be passed through like

{
  "id": {"type_name": "string"},
  "click_ts": {"type_name": "timestamp", "type_args": {"unit": "ms"}},
  "some_value": {"type_name": "decimal128", "type_args": {"precision": 7, "index": 3}},
  ...
}
Or flat or otherwise—I don't really care about arrangement, just that it's flexible enough when we need it

joosthooz:
This code has the basic functionality and is able to convert the fanniemae dataset. It can't handle datatypes with parameters yet.

edponce:
Out of curiosity, are there files/datasets where 1MB has not been enough to resolve its schema? I ask because this can be an improvement that Arrow will benefit as well.

alistaire47:
> Out of curiosity, are there files/datasets where 1MB has not been enough to resolve its schema? I ask because this can be an improvement that Arrow will benefit as well.

They definitely exist, and there's no magic number here; at some point you just have to specify a schema. An example is an export from a database table sorted by timestamp where a nullable field was added at some point—values may be entirely null for millions+ of observations even though there's useful data at the end of the dataset.

Especially for interactive work, being able to specify a partial schema just for columns you know are screwy can be handy; [readr::cols()](https://readr.tidyverse.org/reference/cols.html) takes this approach with a default of inferring types for unspecified columns. Obviously specifying a complete schema is _better_, but forcing users to do so can lead to other antipatterns. What we have is good, just maybe a little idealistic.

edponce:
@alistaire47 Thank you for the explanation. A possible solution for file format standardization is to include metadata with a value per column identifying where the first non-null value appears, this way tools can fetch the value directly for schema resolution.

alistaire47:
> A possible solution for file format standardization is to include metadata with a value per column identifying where the first non-null value appears, this way tools can fetch the value directly for schema resolution.

@edponce Ooh that would be great, and handle a lot of problem cases! The remaining issues I've run into have all been in regards to schema params of fields that are difficult to infer from data, e.g. unit/timezone for timestamps. I think at least timestamp unit handling may be improved (i.e. they roundtrip file <-> memory correctly with defaults) in more recent parquet versions, though?

joosthooz:
@edponce thanks for contributing to the discussion! Suppose you go through the trouble of storing the row number where the first value is encountered, wouldn't it be just as much work to simply store the datatype of the column?

joosthooz:
(Related JIRA BTW: https://issues.apache.org/jira/browse/ARROW-8221)

edponce:
@joosthooz yeah, you are right. There are many variations data details can be provided to help with schema resolution. In an ideal world, most data formats in file/memory would have metadata (min, max, null count, nrows, ncols, types, etc) and consistent across formats.

Another issue which seems more related to this PR's work, revolves on how to handle data updates? Viewing data as a data structure that can be modified directly creates many issues. If datasets would only be allowed to be modified via some sort of API, then metadata could be updated accordingly during data updates.

BTW, this is just me thinking out loud. Thanks for the discussion!

jonkeane:
Is it possible to have this PR be based on main instead of the validation branch? I know they might be inter-related in a way that's hard to tease them apart, but it would make reviewing much easier if I could see what the changes are here for schemas and not the validation bits

joosthooz:
@jonkeane I changed the target branch to the validation branch, so now you only see the additional changes specific to this task. That should help. Once we merge #34 we can change the target branch back to main.

datalogistik/util.py
```diff
@@ -432,7 +511,9 @@ def convert_dataset(
         conv_time = time.perf_counter() - conv_start
         log.info("Finished conversion.")
         log.debug(f"conversion took {conv_time:0.2f} s")
-        dataset_info["tables"] = metadata_table_list
+        # Parquet already stores the schema internally
+        if new_format == csv:
```
austin3dickey, line 515:
```suggestion
        if new_format == "csv":
```

datalogistik/util.py
```diff
@@ -318,15 +385,27 @@ def get_dataset(input_file, dataset_info, table_name=None):
...
             co = csv.ConvertOptions(column_types=column_types_trailed)
+        else:  # not a TPC dataset
+            if dataset_info.get("tables"):
+                log.debug("Found schema information in metadata")
+                for table_entry in dataset_info.get("tables"):
+                    if table_name is None or table_entry["table"] == table_name:
+                        schema = get_arrow_schema(table_entry["schema"])
+                        column_names = list(
+                            table_entry["schema"].keys()
+                        )  # TODO: does this really always preserve ordering?
```
austin3dickey, line 400: How will you verify this TODO? It looks to me like it will always preserve ordering

austin3dickey (changes requested):
nice, this is looking good!

datalogistik/cli.py
```diff
@@ -114,68 +143,77 @@ def parse_args_and_get_dataset_info():
...
+            if dataset_source["name"] == opts.dataset:
+                dataset_info = dataset_source
+                break
+        if dataset_info is None and opts.dataset not in tpc_info.tpc_datasets:
+            msg = (
+                f"Dataset '{opts.dataset}' not found in repository or list of supported "
+                "generators.\n\nDatasets found in repository: "
+                f"{[source['name'] for source in dataset_sources]}\nSupported generators: "
+                f"{tpc_info.tpc_datasets}"
+            )
```
jonkeane, line 178: Did anything in here change other than the indent? I don't think so, but the diff isn't particularly easy to tell

datalogistik/cli.py
```diff
@@ -114,68 +143,77 @@ def parse_args_and_get_dataset_info():
...
+            log.error(msg)
+            raise ValueError(msg)
+
+        if "partitioning-nrows" not in dataset_info:
+            dataset_info["partitioning-nrows"] = 0  # Default: no partitioning
+        return (dataset_info, opts)
+    else:
+        msg = "Please specify a command"
+        log.error(msg)
+        raise RuntimeError(msg)
```
jonkeane, line 219: Same here? If there are changes, maybe highlighting them with a comment describing might be helpful

datalogistik/util.py
```diff
@@ -93,6 +95,90 @@ def add_file_listing(metadata, path):
...
+# a file listing.
+def contains_dataset(path):
+    if not path.exists():
+        msg = f"Path '{path}' does not exist"
+        log.error(msg)
+        raise RuntimeError(msg)
+    if not path.is_dir():
+        msg = f"Path '{path}' is not a directory"
+        log.error(msg)
+        raise RuntimeError(msg)
```
jonkeane, line 108: I might be misremembering, but I feel like I've seen similar sets of code a few places for confirming that something is a path in the cache. I wonder if we should factor these out to helper utilities that we can call?

datalogistik/util.py
```diff
@@ -173,59 +266,106 @@ def clean_cache():
...
 
 
+# Create an instance of the pyarrow datatype with the given name
+def arrow_type_function_lookup(function_name):
+    if isinstance(function_name, str):
+        pa_type_func = getattr(pa, function_name)
+        return pa_type_func
+
+    # The argument was not a pyarrow type (maybe a nested structure?)
+    return None
```
jonkeane, line 302: Would you mind adding some tests for this function? I find it super helpful to read through for things like this to get a better idea of what this is trying to do + what the defensiveness in it is trying to prevent

datalogistik/util.py
```diff
@@ -173,59 +266,106 @@ def clean_cache():
...
+
+
+# Convert a given item (string or dict) to the corresponding Arrow datatype
+def arrow_type_from_json(input_type):
+    # Dict of all pyarrow types that take parameters to their possible parameters
+    # and whether they are optional
```
jonkeane, line 308: How is the optionality specified? I don't think I see anything about optionality below, but I must be missing it?

datalogistik/util.py
```diff
@@ -173,59 +266,106 @@ def clean_cache():
...
+    schema = None
+    format = dataset_info["format"]
+    if format == "parquet":
         dataset_read_format = ds.ParquetFileFormat()
-    if dataset_info["format"] == "csv":
+    if format == "csv":
         # defaults
         po = csv.ParseOptions()
-        ro = csv.ReadOptions()
+        ro = csv.ReadOptions()  # autogenerate_column_names=True)
```
jonkeane, line 368: Could you expand on this comment here? I'm not totally sure what it's intending

tests/test_datalogistik.py
```diff
@@ -100,3 +122,54 @@ def test_convert_dataset_csv_to_parquet():
...
+    test_parquet_file = test_filename + ".parquet"
+    test_parquet_file_path = pathlib.Path(path, test_parquet_file)
+    path.mkdir(parents=True, exist_ok=True)
+    dataset_info = {
+        "name": "test_parquet",
+        "url": "convtest.parquet",
+        "format": "parquet",
+        "partitioning-nrows": 0,
+    }
+    pydict = {"int": [1, 2], "str": ["a", "b"]}
```
jonkeane, line 141: Would it be helpful to have a dataset with one of each type of class and reuse that in tests like this?

tests/test_datalogistik.py
```diff
@@ -100,3 +122,54 @@ def test_convert_dataset_csv_to_parquet():
...
+    print(written_table.schema)
+    assert written_table == orig_table
+    converted_path = util.convert_dataset(dataset_info, None, "parquet", "csv", 0, 0)
+    test_csv_file_path = pathlib.Path(converted_path, test_csv_file)
+    converted_table = ds.dataset(
+        test_csv_file_path, format=ds.CsvFileFormat()
+    ).to_table()
+    print(converted_table.schema)
+    assert converted_table == orig_table
+    util.prune_cache_entry("test_parquet")
```
jonkeane, line 156: I'm not sure if this test does this (it doesn't look like it does, but maybe I'm missing something!), but could we add a test of the actual roundtrip and actually converting schema types? 

It would be good to capture: changing things like int32 to float, int32 to int64, a numeric of some sort to a string, etc. Ideally we would have tests for all of them to confirm that they really do all work (though this should fall into pyarrow pretty quickly, so maybe we don't need total coverage, but at least a decent amount of coverage to have confidence we can do what we say we do would be good)

jonkeane (changes requested):
Thanks for this, this is looking good. I've commented a few places about the validation code that seems to be in here, though I think that code isn't actually intended to be here or be merged in this PR (a rebase might help, or possibly even pulling those commits out totally?).

We should definitely add some more tests in that check various conversion possibilities (and now might be a good time to setup a general smallish dataset we use in our tests generally speaking that contains all or many of the types we expect to see)

joosthooz:
It seems I need to rebase this, I thought retargeting to the main branch after merging #34 would fix it but it didn't.
Thanks for all the comments, especially for all the ideas for tests :+1:

datalogistik/util.py
joosthooz: Thanks for catching this!

datalogistik/util.py
joosthooz: I removed the comment

datalogistik/util.py
joosthooz: I created a TODO in the code here, and also opened an issue about this: https://github.com/conbench/datalogistik/issues/50

datalogistik/util.py
joosthooz: I don't know, it's just that if it didn't, it would become a really unpleasant bug!

README.rst
```diff
@@ -9,7 +9,7 @@ generate some datasets such as TPC-H by calling their (external) generator progr
 
 Usage::
 
-    datalogistik <generate|cache>
+    datalogistik <generate|cache|validate>
```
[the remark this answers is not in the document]
joosthooz: This is outdated code

datalogistik/cli.py
```diff
@@ -96,15 +97,43 @@ def parse_args():
...
+        "--dataset",
+        type=str,
+        help="Path to the dataset to Validate",
+    )
+    valid_group.add_argument(
+        "-c",
+        "--cache",
+        action="store_true",
+        help="Validate all entries in the cache",
+    )
```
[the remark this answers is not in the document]
joosthooz: Also outdated

datalogistik/cli.py
joosthooz: Also outdated

datalogistik/cli.py
joosthooz: Also outdated

joosthooz:
Ok things should be up to date again, I'll get started on the recommendations

tests/test_datalogistik.py
joosthooz: This functionality is not supported yet, and I think it falls outside the scope of the PR. Even though the code for this might even be rather simple, how to interface this to the user will be a challenge! Created a new issue: https://github.com/conbench/datalogistik/issues/58

datalogistik/util.py
joosthooz: I added tests for these new functions

datalogistik/util.py
```diff
@@ -284,16 +287,108 @@ def schema_to_dict(schema):
...
-    if dataset_info["format"] == "csv":
+    if format == "csv":
         # defaults
         po = csv.ParseOptions()
-        ro = csv.ReadOptions()  # autogenerate_column_names=True)
+        ro = csv.ReadOptions()
         co = csv.ConvertOptions()
+        # TODO: Should we autogenerate column names by default?
+        # Or add a property in the metadata about it?
+        # or allow a fall-back to read_csv in case schema detection fails?
```
alistaire47, line 391: We've got both csv datasets with no headers (fanniemae) and with headers (nyctaxi). If we're specifying a schema, we should use the names specified there. If we're not...well, we probably should be

alistaire47: But we should be careful not to accidentally skip the first line if there are no headers and we overwrite ones inferred from the first line of data; we need to store metadata about this for each dataset

datalogistik/util.py
```diff
@@ -284,16 +287,108 @@ def schema_to_dict(schema):
...
+        "halffloat": "float16",
+        "float": "float32",
+        "double": "float64",
+        "decimal": "decimal128",
+    }
+    for (alias, func_name) in aliases.items():
+        if type_name == alias:
+            return func_name
+    # no alias was found
+    return type_name
```
austin3dickey, line 302: nit: you could reduce some complexity here!
```suggestion
    return aliases.get(type_name, default=type_name)
```

datalogistik/util.py
```diff
@@ -284,16 +287,108 @@ def schema_to_dict(schema):
...
+        return arrow_type_function_lookup(type_name)(*args)
+    else:  # args is probably a single value
+        return arrow_type_function_lookup(type_name)(args)
+
+
+# Convert the given dict to a pyarrow.schema
+def get_arrow_schema(input_schema):
+    log.debug("Converting schema to pyarrow.schema...")
+    if input_schema is None:
+        return None
```
austin3dickey, line 364: nit: maybe log after we've checked for null input?
```suggestion
    if input_schema is None:
        return None
    log.debug("Converting schema to pyarrow.schema...")
```

tests/test_datalogistik.py
```diff
@@ -61,13 +69,288 @@ def test_version():
...
+    "o": "date64",
+    "q": "string",
+    "r": "string",
+    "t": "large_string",
+    "u": "large_string",
+    "v": {"type_name": "time32", "arguments": "ms"},
+    "w": {"type_name": "time64", "arguments": "us"},
+    "x": {"type_name": "timestamp", "arguments": {"unit": "ms"}}
+    }"""
+complete_csv_schema_json_output = "{'a': 'null', 'b': 'bool', 'c': 'int8', 'd': 'int16', 'e': 'int32', 'f': 'int64', 'g': 'uint8', 'h': 'uint16', 'i': 'uint32', 'j': 'uint64', 'l': 'float', 'm': 'double', 'n': 'date32[day]', 'o': 'date64[ms]', 'q': 'string', 'r': 'string', 't': 'large_string', 'u': 'large_string', 'v': 'time32[ms]', 'w': 'time64[us]', 'x': 'timestamp[ms]'}"
```
austin3dickey, line 144: Could you please break this long line up? I like this method but you can do whatever works for you:
```
complete_csv_schema_json_output = (
    "string "
    "parts"
)
```

austin3dickey (approved):
This is looking great, especially all these comprehensive tests! Just a few small things.

datalogistik/util.py
joosthooz: Thanks! much better

datalogistik/util.py
joosthooz: Done

tests/test_datalogistik.py
joosthooz: Done. Thanks for pointing it out, I didn't think of this

datalogistik/util.py
```diff
@@ -284,16 +289,112 @@ def schema_to_dict(schema):
...
+    log.debug("Converting schema to pyarrow.schema...")
+    field_list = []
+    # TODO: a `field()` entry is not a (name, type) tuple
+    for (field_name, type) in input_schema.items():
+        log.debug(f"Schema: adding field {field_name}")
+        arrow_type = arrow_type_from_json(type)
+        field_list.append(pa.field(field_name, arrow_type))
+
+    output_schema = pa.schema(field_list)
+    return output_schema
```
alistaire47, line 370: This is super useful. Do we have the reverse (serialize a schema to json) here somewhere as well? (Not sure if it's necessary here at the moment, but I've run into cases where it would've been very handy.) It might be nice to put all this schema stuff in a separate file so when I, um, borrow this code in the future it's in a nice self-contained package.

datalogistik/util.py
```diff
@@ -284,16 +289,112 @@ def schema_to_dict(schema):
...
+        return arrow_type_function_lookup(type_name)()
+    if isinstance(args, Mapping):
+        return arrow_type_function_lookup(type_name)(**args)
+    elif isinstance(args, list):
+        return arrow_type_function_lookup(type_name)(*args)
+    else:  # args is probably a single value
+        return arrow_type_function_lookup(type_name)(args)
+
+
+# Convert the given dict to a pyarrow.schema
```
alistaire47, line 357: If we make all these comments into docstrings they'll pop up nicely for us in editors

datalogistik/util.py
```diff
@@ -383,10 +501,23 @@ def convert_dataset(
                 if parquet_compression is None:
                     parquet_compression = "snappy"  # Use snappy by default
                 write_options = dataset_write_format.make_write_options(
-                    compression=parquet_compression
+                    compression=parquet_compression,
+                    use_deprecated_int96_timestamps=False,
+                    coerce_timestamps="us",
```
alistaire47, line 506: If you've upgraded to arrow 9.0.0, this may be unnecessary now; some of the weird behavior around this was part of the old parquet versions, and the default version got bumped with 9.0.0. Fine to leave here for safety though.

alistaire47 (approved):
A couple nits, but this looks really good!

datalogistik/util.py
joosthooz: I know, that would be really handy and the `inferred-schema` was actually a workaround for not having it at first (but then after thinking about it I think it is good to know the difference between a user-specified schema vs an inferred one).
The other way is more difficult, though. It might come down to a large case statement with all the types with parameters. I remember seeing some Rust code for this at some point, maybe in datafusion, but I couldn't find it anymore.
Let's implement this in the near future.

Side comment; I found this old spec for Arrow Schema JSON representation: https://github.com/apache/arrow/pull/158/files

datalogistik/util.py
joosthooz: Yeah I shouldv'e done that properly from the start. Maybe we should start a PR that does that and adds type hints.

joosthooz: https://github.com/conbench/datalogistik/issues/59

PR #44, opened 2022-08-16
TPC hack cleanups

jonkeane:
This does a few things:

* uses dates instead of timestamps in tpc-h
* uses decimals as the spec states

I know that some engines that we might use this with need hacks to decimal columns, we should allow those as transformations, but supply in-spec data as default.


austin3dickey (approved):
nice!

Issue #17, opened 2022-07-20
Switch to click instead of argparse


austin3dickey:

joosthooz:
I briefly checked out moving to click , but it is slighty more involved than a trivial change because the structure of the program needs to change slightly.
Moreover, what I don't like is that we would lose the argparse parsed options object that carries all of the arguments.
We're passing that around a lot in the code, to save from having to use a dozen of parameters all over the place.
But maybe we can make a function that stores all of the args into a dict, which would be nice because the dataset info is also a dict (read from JSON). That way we don't have to access the args fields with a dot, but the dataset info with [] brackets.

austin3dickey:
I was thinking it would be nice to pass the parameters around as individual variables.
- In my opinion, `def function(scale_factor: int, dataset: str)` is easier to understand than `def function(argument_info: dict)` because it's obvious exactly what inputs it needs, and what inputs it doesn't need.
- A dict loses the ability to use type hints, which are valuable for IDE development and type checking.

That said, I agree that passing ALL the arguments around separately in a `main()` function is cumbersome. So maybe we can pass them as a dict into `main()`, and then whenever we call a util function, that util function takes only the individual arguments it needs rather than the entire `argument_info` dict.

PR #46, opened 2022-08-16
Adding parallel TPC data generation

### LINKED ISSUE #2: Parallelize TPC tools
When generating TPC databases, we should be able to parallelize so that different threads take care of different partitions in the database.

joosthooz:
Resolves #2 
This PR adds code to support directly generating partitioned TPC dataset, and to parallelize TPC data generation.


datalogistik/tpc_builders.py
```diff
@@ -111,13 +115,20 @@ def create_dataset(self, out_dir: pathlib.Path, scale_factor: int = 1):
...
+        for thread in range(partitions):
+            _run(
+                self.executable_path,
+                self.force_flag,
+                self.scale_flag,
+                str(scale_factor),
+                self.partitions_flag,
+                str(partitions),
+                self.current_segment_flag,
+                thread,
```
austin3dickey, line 129:
```suggestion
                str(thread),
```

austin3dickey:
This is looking really interesting so far 👀

joosthooz:
Thanks. I think I covered the basics now. Before these changes, generating partitioned TPC datasets directly was not working properly at all (although conversion was, so you could technically generate a dataset first, and then perform a conversion to get the partitioning you wanted).
So that is added now, and the jobs are run parallel. What I'm going to add now is when generating single-partition TPC datasets, the process is parallelized across files instead of partitions. So it would still be sped up somewhat.
BTW for low scale factors like 1, the speedup is very limited/nonexistent.

datalogistik/tpc_builders.py
```diff
@@ -97,7 +103,20 @@ def __init__(self, executable_path: Optional[pathlib.Path]):
...
+        """Create the pathname for a file in case of a partitioned output"""
+
+    @abc.abstractmethod
+    def _get_table_name_flags(self):
+        """Create a list of the flags needed to generate all the individual tables"""
+
+    def create_dataset(
+        self,
+        out_dir: pathlib.Path,
+        scale_factor: int = 1,
```
jonkeane, line 117: Generally speaking `scale_factor` must be a float (or a decimal, but those are less common as base types). Since we need to support things like `0.1`, `0.01`, etc.

datalogistik/tpc_builders.py
```diff
@@ -106,28 +125,95 @@ def create_dataset(self, out_dir: pathlib.Path, scale_factor: int = 1):
...
+                    f"{table_name}.{self.file_extension}"
+                )
+                new_file = out_dir / (table_name + ".csv")
+                shutil.move(old_file, new_file)
+                # reset permissions to read-only
+                os.chmod(new_file, 0o444)
+                log.debug(f"Created {new_file}")
+        else:
+            num_cpus = multiprocessing.cpu_count()
+            pool_size = num_cpus - 1
```
jonkeane, line 174: Could this | should this be settable by an envvar or an option, so that you can set it to 1 if you want non-parallel, or you could set it to less if you don't want it to use as much, etc. The default here is great when that is unset, though

jonkeane (changes requested):
This is really great, I'm excited to see it in place. 

Since getting this right is quite finicky, and it's easy to have the data produced be subtly wrong, we should definitely have some tests in place for this code confirming that we get the right datasets for low scale factors.

datalogistik/tpc_builders.py
austin3dickey: Ah, I didn't realize you could do a sub-1 SF when I originally type-hinted this code.

joosthooz: Yes, that functionality (non-integer SFs) was lost a very long time ago :sweat_smile: 
We should add it back in at some point. Maybe just tackle this when picking up #17

datalogistik/tpc_builders.py
```diff
@@ -97,7 +103,20 @@ def __init__(self, executable_path: Optional[pathlib.Path]):
...
+
+    @abc.abstractmethod
+    def _get_table_name_flags(self):
+        """Create a list of the flags needed to generate all the individual tables"""
+
+    def create_dataset(
+        self,
+        out_dir: pathlib.Path,
+        scale_factor: int = 1,
+        partitions: int = 1,
```
austin3dickey, line 118: Maybe we should switch the `test-tpc-integration` step to use `-p 100000` or something, so that we can test the crazy parallelization

datalogistik/tpc_builders.py
```diff
@@ -106,28 +125,95 @@ def create_dataset(self, out_dir: pathlib.Path, scale_factor: int = 1):
...
+        else:
+            num_cpus = multiprocessing.cpu_count()
+            pool_size = num_cpus - 1
+
+            with concurrent.futures.ProcessPoolExecutor(pool_size) as pool:
+                futures = []
+                for p in range(1, partitions + 1):
+                    futures.append(
+                        pool.submit(
+                            _run,
```
austin3dickey, line 181: Any reason you didn't use the -parallel flag shipped with dbgen/dsdgen?

austin3dickey (commented):
Looking good! 

I noticed the Windows TPC-DS integration test failed. I'm really hoping it's not a bug with the windows dsdgen compiler because that would be hard to fix. I know it's not super important to support Windows right now, but removing support for something that used to work makes me a little sad. If you can't figure out how to get it to work, maybe we could reserve a piece of the unparallelized code just for windows tpc-ds?

datalogistik/tpc_builders.py
joosthooz: We are - the way it works is that you need to call the tool multiple times in parallel yourself, and use the flags to specify which chunk the current invocation should generate.

datalogistik/tpc_builders.py
joosthooz: One of the difficulties I had in this branch is that the 2 generators seem to treat the partitions flag a bit differently. And because not all tables are suitable for parallel generation (there are only 4 regions and maybe 200 countries). So yeah this makes a lot of sense even only to catch corner cases like these (e.g. a parallelization factor higher than the nr of rows for the given scale factor!)

joosthooz:
The failure in windows CI is a race condition - I'm creating a dictionary of expected metadata that include the current time. After that I call the function that actually generates and writes metadata into a file, as how it would normally work. Sometimes, the clock will just pass a second mark in between these 2 events...

datalogistik/tpc_builders.py
jonkeane: IIRC, at least with the Rust tool for data fusion, having too many partitions (many greater than threads, IIRC) will make the whole thing hang. It would be interested to test this out here too and see if it's also true, and if it is maybe we should add a warning to the docs about that.

joosthooz:
It could actually be faster to generate using a partitioning that allows all CPU threads to be utilized, and then performing a conversion to single partition. Currently when generating single-partitioned data, I use a different code path that parallelizes across tables instead. But for example TPC-H has 1 table that is much larger than the others (it contains like 70% of all the data), so the speedup of that is maybe 40%.
How do you feel about that?

datalogistik/tpc_builders.py
joosthooz: I added an option for this, but that option has the following caveat:
It works to limit the nr of threads used by this code, but not for stuff performed by pyarrow. I've tried setting `pyarrow.set_cpu_count` but it does not seem to work.

datalogistik/tpc_builders.py
joosthooz: A bit hacky rn, because the type of `scale_factor` is now `str` in most places, but I changed it to float here and it works...

jonkeane:
> It could actually be faster to generate using a partitioning that allows all CPU threads to be utilized, and then performing a conversion to single partition

Yeah, this makes a lot of sense. How much work would it take to get there? I think this is a really good way to go: optimize the generation as much as we can and then allow for the output to be as custom as it is without needing to make the tradeoff between generation optimization versus the output form we want.

datalogistik/tpc_builders.py
austin3dickey: I saw Joost made #51 which is great. 

I actually didn't mean to test the limits of `-p` though; I just know that generating TPC-H with a scale factor of 1 means that `-p 100000` gives a reasonable handful of files. So I thought that would be good if we wanted to test a partitioning scheme that resulted in more than 1 file per table, but not too many.

austin3dickey:
>It could actually be faster to generate using a partitioning that allows all CPU threads to be utilized, and then performing a conversion to single partition

Agreed, this would be nice.

>The failure in windows CI is a race condition 
Yikes, I saw this failed a mac unit test too. How might you fix this?

datalogistik/tpc_builders.py
joosthooz: In [d86d316](https://github.com/conbench/datalogistik/pull/46/commits/d86d3160061ded27b08cbd258f7bc108efd926f6) I'm also setting a max for the IO thread pool. Now it seems to be working as expected.

joosthooz:
> How might you fix this?

Should be circumvented by [8c88cdf](https://github.com/conbench/datalogistik/pull/46/commits/8c88cdf7fb6751fe4244bf9b3aed15831fc7e48d)

joosthooz:
> generate using a partitioning that allows all CPU threads to be utilized, and then performing a conversion to single partition

Happy to report that this has been added!

.github/workflows/actions.yml
```diff
@@ -144,5 +144,16 @@ jobs:
         run: pip install poetry
       - name: Create poetry environment
         run: poetry install
-      - name: Run integration test
+      - name: Run TPC-H integration test
```
austin3dickey, line 147: I think it will be cleaner to use pytest's frameworks for managing these various tests. Here's how I might do it:

- Split test files into `tests/unit_tests/test_datalogistik.py`, `tests/integration_tests/test_tpc.py`
- Rename the `unit-tests` GHA job to `test`, and add to it the `msbuild` step from the `test-tpc-integration` job
- Delete the whole `test-tpc-integration` job
- Inside `tests/integration_tests/test_tpc.py`, you can import our python function that generates the dataset (without the CLI), generate it, and run the code that's now in `check_tpc_generation.py`
- Using pytest functionality, you can mark tests that are expected to fail or skip them. I'm also a fan of their `parameterize` decorator that allows you to run the same test multiple times with different params (like maybe `dataset` or `scale_factor` or something)

The main drawback here will be speed; all tests will be run in serial on each OS rather than the TPC code running at the same time on a different runner. But now with parallelized generation that's not as big a deal. Plus, if e.g. Windows always takes longer, we could change the repo settings so that Windows tests don't have to succeed before we merge (for minor PRs, etc).

I also think this setup will still work for developers looking to iterate locally quickly; they could just run `pytest tests/unit_tests` for example. Maybe you could add that to the README.

ci/check_tpc_generation.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def iter_patable_rows(table):
+    for batch in table.to_batches():
+        for row in zip(*batch.columns):
+            yield row
+
+
+failure_occurred = False
+for table in tpc_info.tpc_table_names[dataset]:
```
austin3dickey, line 51: I wonder if we should wrap all this validation code in a function inside `datalogistik` somewhere. Or maybe a method on the TPC builder classes. Not exposing it to users now, but it might be useful to ship it with the package now so it could be used in the future.

ci/check_tpc_generation.py
```diff
@@ -0,0 +1,138 @@
...
+    print(f"validating table {table}...")
+    column_types = tpc_info.col_dicts[dataset][table]
+    column_list = list(column_types.keys())
+    # dsdgen's validation output has a duplicated first column that we need to remove
```
austin3dickey, line 55: haha this is so silly. sorry you had to deal with it.

austin3dickey (changes requested):
Looking good! Looks like the Windows TPC-DS generation is still failing with that message I mentioned earlier. This is happening for every table except `call_center`:
```
ERROR [2022-08-25 14:01:40,578] cmd:
    (WindowsPath('D:/a/datalogistik/datalogistik/datalogistik/dsdgen_tool/tools/dsdgen.exe'), '-FORCE', '-SCALE', '1', '-TABLE', 'store')
ERROR: .\call_center.dat exists. Either remove it or use the FORCE option to overwrite it.
```
It's successfully generating the first table (call_center) but each subsequent table is failing because it's seeing `call_center.dat` already exists, and it's not recognizing the FORCE flag. Not sure why this isn't happening for TPC-H on Windows or other OSes besides that maybe the windows dsdgen compiler doesn't like the way we're parallelizing with it.

Again, supporting Windows TPC-DS is a very low priority, so maybe we just skip that whole test. But it used to work before this PR, so if you have any quick ideas, it might be worth a try.

ci/check_tpc_generation.py
jonkeane: Yup, or at the very least we can include it in the helper files and reuse it there. But that was one of the reasons I also was advocating for having these tests as part of a pytset framework, it's much easier to re-use and modularize bits there than it is having it separated out

ci/check_copyright.py
```diff
@@ -20,6 +20,38 @@
 
 EXCLUSIONS = [
     "poetry.lock",  # autogenerated file
+    "ref_data/tpc-h/0.001/part.tbl",
```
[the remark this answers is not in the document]
jonkeane: Could we copy https://github.com/apache/arrow/blob/cbf0ec0d05fe6301988f3b8f02ea39fead788f6c/dev/archery/archery/utils/rat.py#L45-L48 over and have that working pretty quickly?

datalogistik/config.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import multiprocessing
```
jonkeane, line 15: Is this in the standard lib? I think it is, but want to make sure!

jonkeane (changes requested):
Thanks for keeping up with this. I haven't gone through with a super-fine tooth comb, but some comments. The biggest one is the testing, it would be great to see that folded into our other testing infrastructure here

datalogistik/config.py
joosthooz: It is, there's no extra deps for the poetry project :+1:

datalogistik/tpc_builders.py
```diff
@@ -13,13 +13,17 @@
 # limitations under the License.
 
 import abc
+import concurrent.futures
```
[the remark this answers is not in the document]
joosthooz: Yes!

datalogistik/tpc_builders.py
```diff
@@ -13,13 +13,17 @@
...
 import os
 import pathlib
 import platform
 import shutil
 import subprocess
 from typing import List, Optional
 
+import tqdm
```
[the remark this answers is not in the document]
joosthooz: Ok I'll remove it

ci/check_tpc_generation.py
joosthooz: I moved this code into a new file, `tpc_validation.py`. Now, there's unittests that call the function in that file to validate. For tpc-ds, it is skipped for now.

datalogistik/tpc_builders.py
```diff
@@ -106,28 +131,100 @@ def create_dataset(self, out_dir: pathlib.Path, scale_factor: int = 1):
...
-        for table_name in self.table_names:
-            old_file = self.executable_path.parent / (table_name + self.file_extension)
-            new_file = out_dir / (table_name + ".csv")
-            shutil.move(old_file, new_file)
-            # reset permissions to read-only
-            os.chmod(new_file, 0o444)
-            log.debug(f"Created {new_file}")
+        partitions = 1 if partitions == 0 else partitions
+        num_cpus = get_thread_count()
+        if partitions == 1:
```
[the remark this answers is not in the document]
joosthooz: Good point, done

ci/check_copyright.py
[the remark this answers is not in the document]
joosthooz: Done, let's see if it works

.github/workflows/actions.yml
joosthooz: I did it slightly differently; I deleted the `test-tpc-integration` but I created 2 separate jobs for the 2 files with pytests. Because the tpc tests need the msbuild stuff and because it allows them to run in parallel.

datalogistik/tpc_builders.py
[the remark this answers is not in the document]
jonkeane: In the code blocks below I see two major behavior differences between them: 

* The flags used in the call
* That the partitioned version uses `directory/part-N.csv` in the output

Am I missing anything? 

I wonder if it would be worth it to condense these down into the same block or factor out some of the code that is common to the two of them. and only do the switch for a small section. For example, if you always used the same `directory/part-N.csv`  naming structure, all the file movement code could be the same, yeah? It would be "just" iterating through a list of one

ci/check_copyright.py
```diff
@@ -14,15 +14,21 @@
...
 import subprocess
 from xml.etree import ElementTree
 
 EXCLUSIONS = [
     "poetry.lock",  # autogenerated file
+    "ref_data/*",
 ]
 
 
+def file_is_excluded(file_path):
```
austin3dickey, line 28: Nice!

ci/check_tpc_generation.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2022, Voltron Data.
```
austin3dickey, line 1: Is this script used anymore? If not, you could probably remove it

datalogistik/tpc_builders.py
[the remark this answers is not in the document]
austin3dickey: Yeah, I like that we use these different strategies for the different possible partitionings, but this code is pretty hard to read with the large spread-out `if/else` statements. Functionalizing part of it would help a lot.

datalogistik/tpc_builders.py
```diff
@@ -106,28 +129,98 @@ def create_dataset(self, out_dir: pathlib.Path, scale_factor: int = 1):
...
+            # Move the new files to out_dir
+            out_dir = pathlib.Path(out_dir).resolve()
+            for table_name in self.table_names:
+                old_file = self.executable_path.parent / (
+                    f"{table_name}.{self.file_extension}"
+                )
+                new_file = out_dir / (table_name + ".csv")
+                shutil.move(old_file, new_file)
+                # reset permissions to read-only
+                os.chmod(new_file, 0o444)
```
austin3dickey, line 184: Will you please replace these `chmod` calls with our new `set_readonly` util?

tests/test_tpc.py
```diff
@@ -0,0 +1,83 @@
...
+
+import pytest
+
+from datalogistik import config, datalogistik, tpc_info, tpc_validation
+
+
+@pytest.mark.parametrize("dataset_name", tpc_info.tpc_datasets)
+# this will exercise both parallelization strategies
+@pytest.mark.parametrize("scale_factor", [1, 2])
+@pytest.mark.parametrize("format", config.supported_formats)
```
austin3dickey, line 28: I think you could hardcode `format="csv"` in both these tests in order to cut the run time in half. Using `parquet` will just generate it to CSV and then convert it to parquet, right? But we already test csv -> parquet conversion elsewhere.

tests/test_tpc.py
```diff
@@ -0,0 +1,83 @@
...
+            ]
+            datalogistik.main()
+            assert e.type == SystemExit
+            assert e.value.code == 0
+
+
+@pytest.mark.parametrize("dataset_name", tpc_info.tpc_datasets)
+@pytest.mark.parametrize("format", config.supported_formats)
+@pytest.mark.parametrize("partitioning", [0, 600000])
+def test_validate_tpc_generation(capsys, dataset_name, format, partitioning):
```
austin3dickey, line 54: I wonder if we should skip this test when we're not on ubuntu. I'm assuming if it generates a valid dataset on one OS it'll work on the others too.

tests/test_tpc.py
```diff
@@ -0,0 +1,83 @@
...
+                "-d",
+                dataset_name,
+                "-s",
+                str(scale_factor),
+                "-f",
+                format,
+                "-p",
+                str(partitioning),
+            ]
+            datalogistik.main()
```
austin3dickey, line 46: So we're still seeing the Windows TPC-DS generation error (see https://github.com/conbench/datalogistik/pull/46#pullrequestreview-1087159847 for my comment). I'm cool if we skip that case for now, and maybe make an Issue to come back to it. In that Issue, could you make a note that it seemed to work before this PR?

austin3dickey (changes requested):
nice, really coming along

tests/test_tpc.py
jonkeane: Yeah, we don't need to parameterize across all possible formats here, since we should have tests elsewhere for things like [csv, parquet, arrow] -> [csv, parquet, arrow].

IMO we should do `parquet` here, since that's the most common way that we will be using TPC data.

tests/test_tpc.py
```diff
@@ -0,0 +1,83 @@
...
+            assert e.type == SystemExit
+            assert e.value.code == 0
+
+
+@pytest.mark.parametrize("dataset_name", tpc_info.tpc_datasets)
+@pytest.mark.parametrize("format", config.supported_formats)
+@pytest.mark.parametrize("partitioning", [0, 600000])
+def test_validate_tpc_generation(capsys, dataset_name, format, partitioning):
+    if dataset_name == "tpc-ds":
+        pytest.skip()
```
jonkeane, line 56: If I'm reading this correctly, we test that we can generate tpc-h and tpc-ds up above (with a number of formats) and then in this block we test that we can generate tpc-h data and also validate that data.

If that's the case, we actually end up testing tpc-h twice, once in the first block (without validation) and then once in the second block (with validation) right? If that's the case, let's not run the tpc-h test up top since it's effectively duplicated down below, yeah?

ci/check_copyright.py
joosthooz: Thanks, I adapted it from the snippet that Jon linked :wink:

datalogistik/tpc_builders.py
joosthooz: I tried that but it resulted in a circular dependency... also if I only import the functions

tests/test_tpc.py
joosthooz: Yes parquet includes generation + conversion, and then let's test both formats with the validation, for a very small sf.

tests/test_tpc.py
joosthooz: ok, and again we could keep all platforms for the validation test

tests/test_tpc.py
```diff
@@ -0,0 +1,83 @@
...
+import tempfile
+
+import pytest
+
+from datalogistik import config, datalogistik, tpc_info, tpc_validation
+
+
+@pytest.mark.parametrize("dataset_name", tpc_info.tpc_datasets)
+# this will exercise both parallelization strategies
+@pytest.mark.parametrize("scale_factor", [1, 2])
```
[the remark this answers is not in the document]
joosthooz: The reason for these 2 is that the parallelization strategy is different for sf larger than 1. But I'll trim down the parameters here, as discussed above.

tests/test_tpc.py
joosthooz: Yes, will do

jonkeane:
I spent some time digging into this this afternoon and I ended up creating a branch so I could play around with various ways this works. I noticed a number of things that we can simplify. Which I implemented a handful of in this branch. You can see my changes at https://github.com/conbench/datalogistik/compare/2-parallelize-tpc-tools...2-parallelize-tpc-tools-jtk (feel free to cherry pick those commits directly, I didn't want to _just_ push to your branch with a bunch of changes like that, but IMO we should head towards that direction...)

A few themes:

* I changed the way partitioning works so that when we generate the raw TPC data, we always partition that by the number of threads available. Then the partitioning setting only impacts the _output_ format. [I also did a bit of experimentation and found something of a sweet spot[(https://github.com/conbench/datalogistik/compare/2-parallelize-tpc-tools...2-parallelize-tpc-tools-jtk#diff-645a250cbbb80ba5a037b79a484e49f3ac4ca7aeb78379b5d03c66a8473381a5R733-R753). I don't imagine that anyone really will want to vary how many partitions the raw data is created with: all they care about is the output. So we should generate the raw data in an ideal way for us and then the output gets the partitions requested.
* I changed the type of the raw TPC data since it's not really a csv (as we see with all the code we need to futz with it...). We don't (yet) actually need to generate proper csvs of the tpc data so let's skip that feature for now. The only format we need right now is parquet
* I also added a bit to get a reasonable row group size (turns out, this actually impacts performance quite a bit!). [This code](https://github.com/conbench/datalogistik/compare/2-parallelize-tpc-tools...2-parallelize-tpc-tools-jtk#diff-645a250cbbb80ba5a037b79a484e49f3ac4ca7aeb78379b5d03c66a8473381a5R633-R646) should almost certainly be factored out to a function that we can test it more easily
* I found a number of the more monolithic functions with lots of if/else branches a bit hard to reason about, if possible we should try and break these down into smaller functions that are easier to reason about + test. Here's [one example](https://github.com/conbench/datalogistik/compare/2-parallelize-tpc-tools...2-parallelize-tpc-tools-jtk#diff-645a250cbbb80ba5a037b79a484e49f3ac4ca7aeb78379b5d03c66a8473381a5R450-R519) which honestly is probably still not quite right, but I found much easier to break it down like this (and then ideally we could test the individual functions with all their inputs much easier. 
* If maintaining the ability to have the output show up in partitioning other than 0/1 is difficult, please feel free to rip it all out of here, we can add it later. Right now there are no examples of things that need partitioned data for the mvp (e.g. arrowbench and arrow benchmarks both do not test those data sources)

Very happy to chat 1:1 about some of this tomorrow morning (well, morning my time) if that's helpful!

tests/test_tpc.py
joosthooz: I kept some of the tests, because I want to trigger the both the parallelization strategies. But maybe we'll remove 1 of those strategies, it just adds more complexity and it only improves the speed a little bit for very small scale factors so it is not so important

PR #48, opened 2022-08-19
Improve dealing with location of cache

joosthooz:
Moved cache from current working directory to user profile

### DIFF
--- a/datalogistik/config.py
+++ b/datalogistik/config.py
@@ -14,12 +14,17 @@
 
 import os
 import pathlib
+import platform
 
 debug = True
 
 default_repo_file = str(pathlib.Path(__file__).parent.parent / "repo.json")
-default_cache_location = "./"
-cache_dir_name = "datalogistik_cache"
+if platform.system() == "Windows":
+    default_cache_location = os.getenv("LOCALAPPDATA")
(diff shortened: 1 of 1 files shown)


austin3dickey (approved):
nice, love it.

datalogistik/config.py
```diff
@@ -14,12 +14,17 @@
...
 
 default_repo_file = str(pathlib.Path(__file__).parent.parent / "repo.json")
-default_cache_location = "./"
-cache_dir_name = "datalogistik_cache"
+if platform.system() == "Windows":
+    default_cache_location = os.getenv("LOCALAPPDATA")
+    cache_dir_name = "datalogistik_cache"
+else:  # Unix (Linux or Mac)
+    default_cache_location = os.getenv("HOME")
+    cache_dir_name = ".datalogistik_cache"
```
jonkeane, line 27: Could we also have an over-ride env var here so that someone can override this and tell datalogistik where to put the cache?

joosthooz: That already existed - this only changes the default (in case that env var is not set). You can set `DATALOGISTIK_CACHE` to specify a different location

PR #49, opened 2022-08-19
Add way to discriminate between compression standards in the cache

### LINKED ISSUE #9: Add compression to csv datasets
This would save lots of disk space for relatively little performance cost.
Also, Arrow can read gzip directly.

(De)compressing CSV files should be a conversion just like #49 adds for parquet. So it should be possible to have both the compressed and uncompressed variants of a CSV dataset in the cache.

### LINKED ISSUE #16: Add way to discriminate between parquet compression standards in the cache
When converting a dataset to parquet, the user can provide a compression standard that is passed to the parquet writer. But downloaded datasets in parquet format already have a certain compression. Currently the compression standard is not in the cache hierarchy naming structure, so it is not possible to distinguish a datasets wrt its parquet compression. We should consider adding this

joosthooz:
Fixes #9, fixes #16
This adds a directory hierarchy level, to denote what internal compression is used.
For downloaded parquet files, it first downloads to a directory `parquetcompression_unknown`, then detects the compression and moves it to the proper location in the cache.
- [x] Parquet
- [x] csv

### DIFF
--- a/datalogistik/cli.py
+++ b/datalogistik/cli.py
@@ -75,7 +75,6 @@ Supported formats: Parquet, csv",
         "-c",
         "--compression",
         type=str,
-        default="snappy",
         help="Internal compression (passed to parquet writer)",
     )
     gen_parser.add_argument(
@@ -171,6 +170,11 @@ def parse_args_and_get_dataset_info():
                 msg = "Compression is only supported for parquet format"
                 log.error(msg)
                 raise ValueError(msg)
+        else:
+            if opts.format == "parquet":
+                # Set default here instead of in the CLI code so we can produce
+                # the error above
+                opts.compression = "snappy"
 
         if opts.scale_factor != "" and opts.dataset not in tpc_info.tpc_datasets:
             msg = "scale-factor is only supported for TPC datasets"
--- a/datalogistik/cli.py
+++ b/datalogistik/cli.py
@@ -75,7 +75,8 @@ Supported formats: Parquet, csv",
         "-c",
         "--compression",
         type=str,
-        help="Internal compression (passed to parquet writer)",
+        help="Compression (for parquet: passed to parquet writer, "
+             "for csv: either None of gz)",
     )
     gen_parser.add_argument(
         "-s",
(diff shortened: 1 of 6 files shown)


tests/test_datalogistik.py
```diff
@@ -131,7 +131,9 @@ def test_convert_dataset_csv_to_parquet():
     written_table = csv.read_csv(test_csv_file_path)
     print(written_table.schema)
     assert written_table == orig_table
-    converted_path = util.convert_dataset(dataset_info, None, "csv", "parquet", 0, 0)
+    converted_path = util.convert_dataset(
+        dataset_info, None, None, "csv", "parquet", 0, 0
+    )
```
austin3dickey, line 136: I assume these tests cover a lot of this new code, correct? Is there value in adding another test that converts between different compression formats?

austin3dickey (commented):
Looking great!

joosthooz:
Now both parquet and csv have a cache directory hierarchy level for compression. Conversion between gzip and uncompressed work both ways, but I'm going to change that into re-downloading the file because it is much faster

alistaire47:
> Conversion between gzip and uncompressed work both ways, but I'm going to change that into re-downloading the file because it is much faster

This is definitely not true for all datasets; some of the taxi variants can take >1h to download but gzip/decompress in minutes. And why wouldn't we [usually] keep the downloaded version around anyway, since that's the source of truth? Or keep a quick-to-read version around (probably parquet-snappy) so the primary cost is just resaving?

datalogistik/cli.py
```diff
@@ -75,7 +75,8 @@ def parse_args():
         "-c",
         "--compression",
         type=str,
-        help="Internal compression (passed to parquet writer)",
+        help="Compression (for parquet: passed to parquet writer, "
+        "for csv: either None of gz)",
```
austin3dickey, line 79:
```suggestion
        "for csv: either None or gz)",
```

datalogistik/cli.py
```diff
@@ -165,11 +166,19 @@ def parse_args_and_get_dataset_info():
...
-            if opts.format != "parquet":
-                msg = "Compression is only supported for parquet format"
-                log.error(msg)
-                raise ValueError(msg)
+        # Defaults
+        if opts.compression is None:
+            if opts.format == "parquet":
+                opts.compression = "snappy"
+            if opts.format == "csv":
+                opts.compression = "gz"
```
austin3dickey, line 174: Should we be using `tar.gz`? I think it's more standard for that to be the file extension, no?

tests/test_datalogistik.py
austin3dickey: It's hard for me to tell from this file, but did you cover compressing/decompressing csvs and parquets in these tests?

Also, perhaps there's value in adding compression to our integration tests?

austin3dickey (changes requested):
Nice, a few small things. I also agree with Ed that we should probably prefer converting a local dataset instead of re-downloading it, even at the cost of speed, because fewer network requests is probably a good idea

joosthooz:
I added some tests, but there's still a job to be done there; to be able to test all kinds of conversions between format, partitioning, and compression, we first need to generate a dataset with only datatypes that both parquet and csv support. I'll make a note of that

austin3dickey (commented):
I see the unit tests are sometimes failing with
```
Differing items:
{'local-creation-date': '2022-09-05T10:40:14+0000'} != {'local-creation-date': '2022-09-05T10:40:13+0000'}
```
Will that be fixed in #46 or can you fix it here?

joosthooz:
> Will that be fixed in https://github.com/conbench/datalogistik/pull/46 or can you fix it here?

Yes, it is fixed in #46

austin3dickey (approved):
LGTM

PR #83, opened 2022-09-19
Write JSON, not a string of escaped JSON

jonkeane:
The biggest thing this does is fix a small issue where `write_metadata()` would write an escaped string as JSON, instead of taking the string and writing that directly.

I also updated the default scale_factor to` None` not `''`

I also cleaned up a number of TODOs that are already taken care of

### DIFF
--- a/datalogistik/cli.py
+++ b/datalogistik/cli.py
@@ -79,7 +79,7 @@ Supported formats: Parquet, csv",
         "-s",
         "--scale-factor",
         type=str,
-        default="",
+        default=None,
         help="Scale factor for TPC datasets",
     )
     # TODO: should this be an env var instead of an argument?
--- a/datalogistik/dataset.py
+++ b/datalogistik/dataset.py
@@ -76,9 +76,6 @@ class Dataset:
     # To be filled in programmatically when a dataset is created
     local_creation_date: Optional[str] = None
 
-    # TODO: probably can be deleted
-    rel_path: Optional[pathlib.Path] = None
-
     # TODO: Post-init validation for things like delim if csv, etc.
 
     def __eq__(self, other):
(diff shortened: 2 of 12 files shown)


datalogistik/cli.py
```diff
@@ -79,7 +79,7 @@ def parse_args():
         "-s",
         "--scale-factor",
         type=str,
```
joosthooz, line 81: We could modify the type to float here, too

joosthooz (approved):

datalogistik/dataset.py
```diff
@@ -379,7 +376,8 @@ def write_metadata(self):
         if metadata_file_path.exists():
             util.set_readwrite(metadata_file_path)
         with open(metadata_file_path, "w") as metadata_file:
-            json.dump(obj=json_string, fp=metadata_file)
+            # TODO: how could we use json.dump(self, metadata_file) while still getting the custom serializer to_json() below?
```
austin3dickey, line 379: I don't think it's necessarily "better" to use `json.dump()` than opening the file directly. If we have a custom serializer, it's fine to open the file.

jonkeane: _nods_ Thoughts @alistaire47 ?

alistaire47: Writing to the file means you've got to hold code (inasmuch as json is code) as strings, which I always try to avoid because it's so easy to screw things up without noticing (here, create unparseable json). So my instinct would make `.to_json()` return a dict or list, not a string, or just make a full-on `.write_json()` method that just takes a path; a string as an intermediary is just a chance to screw something up. That said, it's a fairly straight line from `json.dumps()` to `.write()` right now, so if we just don't want to bother right now, that's ok.

PR #84, opened 2022-09-20
Add interface for getting metadata

### LINKED ISSUE #54: Interface for getting metadata
Frequently in benchmarks we validate results by checking attributes of the results, e.g. that reading a CSV into a dataframe was successful because the number of rows and cols match stored metadata. We don't want to store this metadata in benchmarks; it should be parameterized with the data source. It does exist in datalogistik metadata files, but are we ok with benchmarks parsing those? If not, I think we need a command to access metadata, e.g. `datalogistik metadata type_floats` or whatever. Mostly we'll just want dims, but maybe sometimes schemas or other metadata too? So probably a JSON return if we do want a command.

joosthooz:
Closes #54
Adds a `metadata` command that prints out all the metadata of a dataset instance in the cache. If it does not exist, it will _not_ instantiate it and produce an error.

### DIFF
--- a/datalogistik/cli.py
+++ b/datalogistik/cli.py
@@ -17,6 +17,7 @@ import sys
 
 from . import util
 from .dataset import Dataset
+from .dataset_search import find_exact_dataset
 from .log import log
 
 
@@ -28,6 +29,7 @@ def parse_args():
     sub_parsers = parser.add_subparsers(dest="command")
     cache_parser = sub_parsers.add_parser("cache")
     gen_parser = sub_parsers.add_parser("get")
+    meta_parser = sub_parsers.add_parser("metadata")
 
     cache_group = cache_parser.add_mutually_exclusive_group()
     cache_group.add_argument(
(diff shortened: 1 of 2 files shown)


datalogistik/cli.py
```diff
@@ -94,6 +96,36 @@ def parse_args():
...
+        required=True,
+        help="Name of the dataset",
+    )
+    meta_parser.add_argument(
+        "-f",
+        "--format",
+        type=str,
+        required=True,
+        help="Format for the dataset. \
+Supported formats: Parquet, csv",
```
austin3dickey, line 112:
```suggestion
Supported formats: parquet, csv",
```

austin3dickey (commented):
looks good! one small thing

.github/workflows/actions.yml
```diff
@@ -123,5 +123,5 @@ jobs:
       - name: Install package
         run: ${{ matrix.installer }} install '${{ matrix.installee }}'
       - name: Test downloading and converting a dataset
-        run: datalogistik generate -d type_floats -f csv
+        run: datalogistik get -d type_floats -f csv
```
jonkeane, line 126: 😍

jonkeane (commented):
Thanks for all of these cleanups! Lots of silly typos from me in there and very happy to see those fixed. Also I do like the shorter + slightly more evocative `get`

Could we also think about the user-exposed metadata function and if it helps us get to the MVP? I know that @alistaire47 requested this, maybe he could provide some use-cases where we need that for getting this working with arrowbench + labs/benchmarks. But IIUC, we need `get` to be exposed and that's it. 

I know I'm pushing hard on adding only what we need for the MVP and only exposing to end users what we must expose, not what might be helpful. But doing that will let us keep the amount of code pathways we will have to support low + keep us focused on getting to the MVP point.

joosthooz:
In that case, we can leave or close this PR and just use `get`. We might consider an option to output all of the metadata, or just show the user in the docs where they can find the metadata file from the output.
I separated out the other changes here: https://github.com/conbench/datalogistik/pull/88

PR #87, opened 2022-09-20
Add dims to the output result

### LINKED ISSUE #77: Add `dims` to the output result
Since many downstream consumers will want to know the dimensions we truly expect for output validation, we should provide that in the JSON as well.

joosthooz:
Closes #77

### DIFF
--- a/datalogistik/dataset.py
+++ b/datalogistik/dataset.py
@@ -75,7 +75,6 @@ class Dataset:
     # To be filled in programmatically when a dataset is created
     local_creation_date: Optional[str] = None
 
-    # TODO: Post-init validation for things like delim if csv, etc.
     def __post_init__(self):
         if self.format is not None and self.format not in config.supported_formats:
             msg = f"Unsupported format: {self.format}. Supported formats: {config.supported_formats}"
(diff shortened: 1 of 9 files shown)


datalogistik/dataset.py
```diff
@@ -67,6 +67,8 @@ class Dataset:
     metadata_file: Optional[pathlib.Path] = None
     url: Optional[str] = None
     homepage: Optional[str] = None
+    # TODO: should dim actually be at the Table level?
+    # until it is, its entries should correspond to the entries in the tables list
```
jonkeane, line 71: 💯  dims should be at the table level

joosthooz: And for the output, should we make the table entries more complex or create 2 separate lists that correspond to each other?

jonkeane (changes requested):
This looks great — agreed with your comment that this should actually be at the table level. Would you mind adding it there in this PR?

datalogistik/dataset.py
jonkeane: Definitely make the table entries more complex. Any downstream consumer will need to iterate through that list anyway, so they will be able to pull stuff out, no reason to have a second one that needs to be iterated through separately

datalogistik/dataset.py
```diff
@@ -77,6 +76,13 @@ class Dataset:
     local_creation_date: Optional[str] = None
 
     # TODO: Post-init validation for things like delim if csv, etc.
```
jonkeane, line 78:
```suggestion
```

datalogistik/dataset.py
```diff
@@ -509,12 +515,24 @@ def convert(self, new_dataset):
...
         output = {"name": self.name, "format": self.format}
 
         tables = {}
         for table in self.tables:
-            tables[table.table] = str(self.ensure_table_loc(table))
+            table_entry = {}
+            if table.dim:
+                table_entry["dim"] = table.dim
+            table_entry["path"] = str(self.ensure_table_loc(table))
+            tables[table.table] = table_entry
```
jonkeane, line 526: This is slightly more simple (it also always has `dim` which IMO is good for an API like this since other things will parse this output and then we can always rely on it being there

```suggestion
            tables[table.table] = {
                "dim":  table.dim, # might need if table.dim after this to set to None
                "path": str(self.ensure_table_loc(table))
            }
```

datalogistik/dataset.py
```diff
@@ -509,12 +515,24 @@ def convert(self, new_dataset):
...
         output["tables"] = tables
 
         return json.dumps(output)
+
+    def fill_in_nones(self, other_dataset_info):
```
jonkeane, line 531: A small docstring here might be nice to describe what this is doing.

Also, maybe we could call `other_dataset_info` something like `dataset_for_default` or something so it's clear(er) that's where defaults are coming from

jonkeane (approved):
Looking good, a few relatively minor comments | suggestions. Thanks!

PR #90, opened 2022-09-21
Allow `compress` and `decompressed to uncompressed == "none" == None

jonkeane:
Another small change, this for tor the compress|decompress functions to accept our various spellings of uncompressed.


joosthooz:
We could also do this when init'ing a Dataset class (or just after CLI parsing), for example here https://github.com/conbench/datalogistik/pull/87/files#diff-e2e1d83f53d914c8964718cbed8ac9d36bfb28857b1851172331435b41b921bbR79

PR #91, opened 2022-09-21
Clean up a path bug + add tests for it

jonkeane:
The copy-to-temp step had a small bug where it was repeating the path twice in some circumstances. This addresses that + adds two integration tests that confirm that at least one pass through these will succeed.

### DIFF
--- a/datalogistik/datalogistik.py
+++ b/datalogistik/datalogistik.py
@@ -31,6 +31,9 @@ def finish():
 
 
 def main(dataset=None):
+    # dataset here should typically be None, so then we use parse_args_and_get_dataset_info() to
+    # create the dataset to use. But it can be helpful in tests to construct ones own dataset
+    # with Dataset(name="my dataset", format="very_fancy") and pass it as the dataset argument
     if dataset is None:
         dataset = cli.parse_args_and_get_dataset_info()
 
--- a/datalogistik/datalogistik.py
+++ b/datalogistik/datalogistik.py
@@ -30,7 +30,7 @@ def finish():
     sys.exit(0)
 
 
-def main(dataset):
+def main(dataset=cli.parse_args_and_get_dataset_info()):
     if config.get_max_cpu_count() != 0:
         pyarrow.set_cpu_count(config.get_max_cpu_count())
         pyarrow.set_io_thread_count(config.get_max_cpu_count())
--- a/datalogistik/datalogistik.py
+++ b/datalogistik/datalogistik.py
@@ -30,9 +30,7 @@ def finish():
     sys.exit(0)
 
 
-def main():
-    dataset = cli.parse_args_and_get_dataset_info()
-
+def main(dataset):
     if config.get_max_cpu_count() != 0:
         pyarrow.set_cpu_count(config.get_max_cpu_count())
         pyarrow.set_io_thread_count(config.get_max_cpu_count())
@@ -58,4 +56,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    main(dataset=cli.parse_args_and_get_dataset_info())
--- a/datalogistik/datalogistik.py
+++ b/datalogistik/datalogistik.py
@@ -30,7 +30,10 @@ def finish():
     sys.exit(0)
 
 
-def main(dataset=cli.parse_args_and_get_dataset_info()):
+def main(dataset=None):
+    if dataset is None:
+        dataset = cli.parse_args_and_get_dataset_info()
+
     if config.get_max_cpu_count() != 0:
         pyarrow.set_cpu_count(config.get_max_cpu_count())
         pyarrow.set_io_thread_count(config.get_max_cpu_count())
@@ -56,4 +59,4 @@ def main(dataset=cli.parse_args_and_get_dataset_info()):
 
 
 if __name__ == "__main__":
-    main(dataset=cli.parse_args_and_get_dataset_info())
+    main()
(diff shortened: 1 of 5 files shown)


datalogistik/datalogistik.py
```diff
@@ -30,8 +30,9 @@ def finish():
     sys.exit(0)
 
 
-def main():
-    dataset = cli.parse_args_and_get_dataset_info()
+def main(dataset=None):
+    if dataset is None:
+        dataset = cli.parse_args_and_get_dataset_info()
```
jonkeane, line 35: This makes it easier to test with various dataset objects

tests/test_datalogistik.py
```diff
@@ -330,3 +337,46 @@ def test_compress(comp_string):
...
+
+        captured = json.loads(capsys.readouterr().out)
+        assert captured["name"] == "fanniemae_sample"
+        assert captured["format"] == "parquet"
+        assert isinstance(captured["tables"], dict)
+
+    finally:
+        # cleanup all files that weren't there to start with
+        for file in set(os.listdir(test_dir_path)) - set(start_files):
+            shutil.rmtree(pathlib.Path(test_dir_path, file))
```
jonkeane, line 382: Is this the right pattern for cleaning up a messy test?

tests/test_tpc.py
```diff
@@ -41,7 +41,7 @@ def test_validate_tpc_generation(capsys, dataset_name, format):
                 "-f",
                 format,
             ]
-            datalogistik.main()
+            datalogistik.main(dataset=cli.parse_args_and_get_dataset_info())
```
jonkeane, line 44:
```suggestion
            datalogistik.main()
```

datalogistik/dataset.py
```diff
@@ -480,9 +480,7 @@ def convert(self, new_dataset):
                     # Convert from name.format/part-0.format to simply a file name.format
                     # To stay consistent with downloaded/generated datasets (without partitioning)
                     # TODO: do we want to change this in accordance to tpc-raw?
-                    tmp_dir_name = pathlib.Path(
-                        output_file.parent, f"{output_file}.tmp"
-                    )
+                    tmp_dir_name = pathlib.Path(f"{output_file}.tmp")
```
alistaire47, line 483: Not quite sure what's going on here, but this is now fully relative, which is may not have been before. Are we sure there aren't situations in which this will write to a directory other than what's expected?

tests/test_datalogistik.py
alistaire47: You could put things in a tempdir with `tempfile.TemporaryDirectory()` or whatever just in case the cleanup doesn't get run for whatever reason, but this seems pretty safe

austin3dickey: I kinda like using a [TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#examples) which I know will be cleaned up no matter what. Then you don't need the try-finally.

tests/test_tpc.py
```diff
@@ -19,7 +19,7 @@
 
 import pytest
 
-from datalogistik import datalogistik, tpc_validation
+from datalogistik import cli, datalogistik, tpc_validation
```
alistaire47: Either this was failing before or this is unused, no?

tests/test_datalogistik.py
austin3dickey: also, this will fix the failing windows test haha

jonkeane: _nods_ yeah that works generally, though here we're operating inside of the datalogistik cache.

I guess we could copy the cache fixture to the temp directory and then we know it's gone... I'll make an issue for that

datalogistik/datalogistik.py
```diff
@@ -30,8 +30,9 @@ def finish():
     sys.exit(0)
 
 
-def main():
-    dataset = cli.parse_args_and_get_dataset_info()
+def main(dataset=None):
```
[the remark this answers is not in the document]
alistaire47: eh, after reading I guess this is a `Dataset` instance, but more specifically the metadata for one that's for the right dataset, but not necessarily the right format. So a type hint wouldn't do it anyway; it'd have to be a comment or docstring

alistaire47 (approved):
Looks good, just a couple questions

datalogistik/dataset.py
jonkeane: This was the bug I was correcting. It was actually still relative before, but then also adding the parent on to it. So if output_file was like `foo/bar/file.parquet` this made `tmp_dir_name` into `/User/name/.datalogistik_cache/foo/bar/foo/bar/file.parquet` which didn't exist and 💣

tests/test_tpc.py
jonkeane: I'm not totally sure why flake8 in CLI complained only now about this, but it did

datalogistik/datalogistik.py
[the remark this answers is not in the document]
jonkeane: Done

datalogistik/dataset.py
alistaire47: Ah. I'm still minorly concerned about the relative-ness going wrong if this gets called from somewhere else; could you add another `.parent` or make this relative to the root dir of the cache?

tests/test_datalogistik.py
jonkeane: https://github.com/conbench/datalogistik/issues/92

datalogistik/dataset.py
jonkeane: #89 is working on restructuring the path getters (which is where this comes from in the first place). Let's address how those paths are treated systematically there, since it'll all change over there anyway (and will catch other possible places that we might need to be paranoid)

PR #93, opened 2022-09-22
default to none compression

jonkeane:
This happens at the init stage for now so that CLI calls go from "uncompressed" -> None. 

In the future we might consider making this a bit less funky: using `None` meaningfully here is a bit odd compared to it being unspecified elsewhere... but the compression argument in pyarrow doesn't take `uncompressed` it takes `None`, so we would need to do a whole getter | setter thing for compression where we store "uncompressed" for all of them but then when we get it for writing we swap to `None`. Not sure if it's worth all of that, so let's kick that decision down the road.

### DIFF
--- a/datalogistik/dataset.py
+++ b/datalogistik/dataset.py
@@ -83,6 +83,15 @@ class Dataset:
         if self.scale_factor is None and self.name in tpc_info.tpc_datasets:
             self.scale_factor = 1.0
 
+        # Use None as the true default for uncompressed
+        # the first comparisson is a bit redundant, but None.lower() fails
+        if (
+            self.compression is None
+            or self.compression.lower() == "none"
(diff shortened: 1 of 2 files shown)


austin3dickey (approved):

joosthooz:
This causes problems for me because I want to use the compression in the repo file if it is unspecified on the cmdline. But after these changes it is no longer clear whether nothing was specified or if the user gave an uncompressed flag

jonkeane:
> This causes problems for me because I want to use the compression in the repo file if it is unspecified on the cmdline

I don't think we really want to reference the repo file as the default see more discussion at https://github.com/conbench/datalogistik/pull/95#discussion_r978621118 for some thoughts on that

joosthooz:
Ok then let's continue that discussion over there.
But this change I've reverted in #89

PR #95, opened 2022-09-23
Handle csv.gz extension, extra null strings, and download tiles to table names

jonkeane:
Sorry for the handful of different changes here, ran into these sequentially getting this to work with arrowbench.

* handles the .csv.gz suffix (this was the minimal possible way to support it, we might want to come back and make an extension helper method to help do that more categorically)
* adds an "extra_nulls" field to datasets that let us specify strings to treat as nulls.
  This is mostly needed for the nyctaxi csv which has a `*` in an otherwise numeric column. The odd thing is that `pyarrow.csv.read_csv()` handles this case _just fine_ but `pyarrow.dataset()` does not. I'm not totally sure we want to extend out the repo + datasets arguments quite like this in the long run, but given that everything can default to null I vote let's do this for now and see if it grows or becomes more of a problem.
* uses "gzip" internally, but accepts "gz" or "gzip"
* Ensures that a download gets saved as `{table_name}.{format}` instead of whatever it's stored as (this is most prominent in our current taxi dataset)
* A small renaming of a variable in the cli to make things extra extra clear what the dataset to fill defaults from is
* A bunch of additional tests for this functionality + some untested functionality around this

### DIFF
--- a/datalogistik/cli.py
+++ b/datalogistik/cli.py
@@ -135,9 +135,9 @@ def parse_args_and_get_dataset_info():
         # TODO:
         #  * compression (in particular: test supported file-compression)
         #  * partitioning (later)
-        dataset_info = repo.search_repo(opts.dataset, repo.get_repo())
-        if dataset_info:
-            dataset.fill_in_defaults(dataset_info)
+        dataset_from_repo = repo.search_repo(opts.dataset, repo.get_repo())
+        if dataset_from_repo:
+            dataset.fill_in_defaults(dataset_from_repo)
         if dataset.compression is None and dataset.format == "parquet":
             dataset.compression = "snappy"
 
--- a/datalogistik/dataset.py
+++ b/datalogistik/dataset.py
@@ -67,6 +67,8 @@ class Dataset:
     metadata_file: Optional[pathlib.Path] = None
     url: Optional[str] = None
     homepage: Optional[str] = None
+    # a list of strings that can be added when csv parsing to treat as if they were nulls
+    extra_nulls: Optional[List] = field(default_factory=list)
 
     # To be filled in at run time only
     cache_location: Optional[pathlib.Path] = None
@@ -91,6 +93,10 @@ class Dataset:
             or self.compression.lower() == "uncompressed"
         ):
             self.compression = None
+
+        # munge gz to gzip
+        if self.compression is not None and self.compression.lower().startswith("gz"):
+            self.compression = "gzip"
 
     def __eq__(self, other):
         if not isinstance(other, Dataset):
@@ -166,7 +172,11 @@ class Dataset:
         if len(table.files) > 1 or table.multi_file:
             name = name
         else:
-            name = name + os.extsep + self.format
+            ext_string = self.format
+            # custom extension for the special case .csv.gz
+            if self.format == "csv" and self.compression == "gzip":
+                ext_string = ext_string + os.extsep + "gz"
+            name = name + os.extsep + ext_string
 
         return name
 
@@ -239,6 +249,10 @@ class Dataset:
         if self.delim:
             po = csv.ParseOptions(delimiter=self.delim)
 
+        # add extra nulls
+        if self.extra_nulls:
+            co.null_values = co.null_values + self.extra_nulls
+
         column_names = None
         # TODO: Where does this go?
         # autogen_column_names = False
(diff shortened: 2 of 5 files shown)


datalogistik/cli.py
```diff
@@ -135,9 +135,9 @@ def parse_args_and_get_dataset_info():
         # TODO:
         #  * compression (in particular: test supported file-compression)
         #  * partitioning (later)
-        dataset_info = repo.search_repo(opts.dataset, repo.get_repo())
-        if dataset_info:
-            dataset.fill_in_defaults(dataset_info)
+        dataset_from_repo = repo.search_repo(opts.dataset, repo.get_repo())
+        if dataset_from_repo:
+            dataset.fill_in_defaults(dataset_from_repo)
```
jonkeane, line 140: This isn't a _huge deal _ but though tit would help our future selves to remember what `dataset_from_repo` was doing there to differentiate it from the more generic `dataset`

datalogistik/dataset.py
```diff
@@ -166,7 +172,11 @@ def get_table_name(self, table):
         if len(table.files) > 1 or table.multi_file:
             name = name
         else:
-            name = name + os.extsep + self.format
+            ext_string = self.format
+            # custom extension for the special case .csv.gz
+            if self.format == "csv" and self.compression == "gzip":
+                ext_string = ext_string + os.extsep + "gz"
+            name = name + os.extsep + ext_string
```
jonkeane, line 179: We might want to factor all this stuff out into an extension method that would let us read format + compression etc. and produce the right extension, but for now this is ok + works

datalogistik/dataset.py
```diff
@@ -540,5 +560,6 @@ def fill_in_defaults(self, dataset_for_defaults):
             if (
                 getattr(self, attr) is None
                 and getattr(dataset_for_defaults, attr) is not None
+                and attr != "compression"
```
jonkeane, line 563: This turns off default compression filling in. I'm pretty sure that's right, since if I request `dl get -d some_dataset -f parquet` I don't know (and shouldn't need to care about) what the default compression is in the repo, I'm telling it there to use the default for that format (which means we should decide what the default is for each format, if that's uncompressed, etc.). But if we fall back to the repo there, the clients would need to know about what's in the repo to make a correct call (or I guess always provide the compression they care about...)

tests/test_dataset.py
```diff
@@ -270,7 +281,7 @@ def test_output_result():
                             "test_cache",
                             "chi_taxi",
                             "dabb1e5",
-                            "chi_traffic_sample.csv",
+                            "chi_traffic_sample.csv.gz",
```
jonkeane, line 284: This change is needed for the test to pass with this new setup, but it would have actually failed in real life since this file is stored at `chi_traffic_sample.csv.gz` but we were sending `chi_traffic_sample.csv` 🙈 oops!

tests/test_dataset.py
```diff
@@ -326,6 +337,39 @@ def _fake_generate(dataset):
...
+        name="tester",
+        format="csv",
+        extra_nulls=["FANCY_NULL"],
+        tables=[Table(table="foo")],
+    )
+    spec, schema = ds.get_csv_dataset_spec(ds.tables[0])
+    # it's a pyarrow quirk that the convert_options are under `default_fragment_scan_options`
+    assert (
+        "FANCY_NULL" in spec.default_fragment_scan_options.convert_options.null_values
+    )
```
jonkeane, line 351: I put this in the comment here too, but I spent _way too long_ figuring out how to access the `convert_options` from a spec object like this. It took RTFSing pyarrow to figure out that `default_fragment_scan_options` isn't _actually_ "default" per se when given read or convert options, and those end up in that attribute there...

joosthooz: I know exactly what you mean, I spent a lot of time on that code for an Arrow PR recently...

tests/test_dataset.py
joosthooz: Yes I noticed that at some point, nice to have it resolved

datalogistik/dataset.py
joosthooz: Isn't it just the other way around? If the user doesn't specify anything I think it would be nice to default to whatever was in the repo. Except of course for parquet, because then the default should just be snappy.
Right now it is not such of a problem, because we only support gz and all files in the repo are gz compressed. But if we have a dataset that is bz2/lz4 compressed, a `get` command would trigger a download and a conversion...

datalogistik/dataset.py
joosthooz: Agreed. I created a helper function for this in #89

datalogistik/cli.py
joosthooz: Yes, this is much better

joosthooz (approved):
This has a small amount of overlap with #89 , because that also fixes the problem with .gz extension, but not too serious to prevent merging this.

PR #97, opened 2022-09-26
Fix clean, prune, and validation commands

joosthooz:
Reworked CLI a bit, so validation, pruning and cleaning work again.

### DIFF
--- a/datalogistik/cli.py
+++ b/datalogistik/cli.py
@@ -129,13 +129,6 @@ def parse_args_and_get_dataset_info():
         compression=opts.compression,
     )
 
-    # Set defaults and perform sanity-check for the arguments:
-    # TODO:
-    #  * compression (in particular: test supported file-compression)
-    #  * partitioning (later)
-    dataset_from_repo = repo.search_repo(opts.dataset, repo.get_repo())
-    if dataset_from_repo:
-        dataset.fill_in_defaults(dataset_from_repo)
     if dataset.compression is None and dataset.format == "parquet":
         dataset.compression = "snappy"
 
@@ -143,6 +136,13 @@ def parse_args_and_get_dataset_info():
     opts.partition_max_rows = 0
 
     if opts.command == "get":
+        # Set defaults and perform sanity-check for the arguments:
+        # TODO:
+        #  * compression (in particular: test supported file-compression)
+        #  * partitioning (later)
+        dataset_from_repo = repo.search_repo(opts.dataset, repo.get_repo())
+        if dataset_from_repo:
+            dataset.fill_in_defaults(dataset_from_repo)
         return dataset
 
     entry = dataset_search.find_exact_dataset(dataset)
--- a/datalogistik/cli.py
+++ b/datalogistik/cli.py
@@ -97,31 +97,74 @@ def parse_args():
     return parser.parse_args()
 
 
-def parse_args_and_get_dataset_info():
+def list_cache_entries(opts):
+    if opts.repo:
+        log.info(
+            "Datasets found in repository: "
+            f"{[source.name for source in repo]}\nSupported generators: "
+            f"{tpc_info.tpc_datasets}"
+        )
+    else:  # assume --cache
+        log.info("Dataset entries currently present in cache:")
+        for dataset in config.get_cache_location().iterdir():
+            metadata_files = config.get_cache_location().glob(
+                f"**/{dataset.name}/**/{config.metadata_filename}"
+            )
+
+            for entry in [Dataset.from_json(ds) for ds in metadata_files]:
+                log.info(entry)
+
+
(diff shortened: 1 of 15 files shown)


joosthooz:
This is a rather large restructure of the CLI. I didn't move over to using click (i.e. #17 ), because I don't think I can assign an argument to multiple subparsers as I'm doing right now with argparse.

tests/test_dataset.py
```diff
@@ -367,7 +367,10 @@ def test_fill_in_defaults():
 
     ds.fill_in_defaults(dataset_from_repo)
     assert ds.format == "parquet"  # NB: not csv
-    assert ds.compression is None
+    if ds.format == "parquet":
+        assert ds.compression == "snappy"
+    else:
+        assert ds.compression is None
```
jonkeane, line 373:
```suggestion
    assert ds.compression == "snappy"
```

This is the only path that would ever be taken, no?

datalogistik/util.py
```diff
@@ -151,8 +151,22 @@ def validate(path):
         raise RuntimeError(msg)
     metadata_file = pathlib.Path(path, config.metadata_filename)
     with open(metadata_file) as f:
-        orig_file_listing = json.load(f).get("files")
-    dataset_valid = validate_files(path, orig_file_listing)
+        tables = json.load(f).get("tables")
```
jonkeane, line 154: If we're touching this code, could you please update it to use the `Dataset` and `Table` path ways? That should allow us to get rid of some of the boiler plate stuff here

datalogistik/dataset_search.py
```diff
@@ -51,6 +52,16 @@ def find_or_instantiate_close_dataset(dataset):
...
 
+    if variants == []:
+        msg = (
+            f"Dataset '{dataset.name}' not found in repository or list of supported "
+            "generators.\n\nDatasets found in repository: "
+            f"{[source.name for source in repo.get_repo()]}\nSupported generators: "
+            f"{tpc_info.tpc_datasets}"
+        )
+        log.error(msg)
+        raise ValueError(msg)
```
jonkeane, line 63: Do we have a test for this behavior? If not could we add one?

datalogistik/cli.py
```diff
@@ -26,62 +26,62 @@ def parse_args():
         description="Dataset cacher/converter and generator",
     )
     sub_parsers = parser.add_subparsers(dest="command")
-    cache_parser = sub_parsers.add_parser("cache")
+    sub_parsers.add_parser("clean")
+    list_parser = sub_parsers.add_parser("list")
     gen_parser = sub_parsers.add_parser("get")
+    prune_parser = sub_parsers.add_parser("prune")
+    val_parser = sub_parsers.add_parser("validate")
```
jonkeane, line 33: Do we have good user stories for how these will be used? 

I can see a good use-case for `clean`, while testing (or to allow clients to blow everything away and restart...). But the others are more speculative IMO. I would be a huge advocate of keeping things hidden from the end-user API until something forces us to expose them. It'll let us be much more quick with our movement and not need to worry as much about changing behaviors

jonkeane (commented):
A few comments, the biggest one is: IMO we should defer exposing things we don't need for the MVP. We want to build what we need to get this being used in the places we've identified (arrowbench + labs/benchmarks) first, and only add public API features if we absolutely have to.

joosthooz:
Sure, we need to focus on the MVP, but currently there is a lot of code that isn't used and doesn't work anymore. So we should remove it or fix it, I feel. So I also fixed the commands to use them. As a middle way we could wait before adding CLI commands for them for now?
The `clean` command checks for dangling dirs that do not have a valid metadata file
`prune` allows the removal of an entry (a single one, currently, it is easy to add a `-a` flag but that felt dangerous)
`validate` checks the files of a dataset agains the checksums in the metadata file. It has options to check all entries, and to remove failing entries

joosthooz:
I created a new PR that doesn't rework the CLI. The clean, validate, and prune-invalid options are still available. But the prune-entry option was removed because it doesn't support the new Dataset class. We can close this PR if we prefer that.

jonkeane:
> So I also fixed the commands to use them. As a middle way we could wait before adding CLI commands for them for now?

💯 agreed on this — making it work + not exposing it (yet) is a good middle way to not lose the code (though we've got it in git too...)

joosthooz:
In the meantime, I'm keeping this open, because I still need to address some of the review comments

datalogistik/util.py
joosthooz: This code has been moved to the Dataset class. The difficulty is that we can't use the Dataset class here because that would introduce a circular dependency...

joosthooz:
All comments should be addressed in #100 so I'm closing this.

PR #99, opened 2022-09-27
Attempt at adding support for downloading an S3 bucket

joosthooz:
I have no way to test this because I don't have the proper AWS credentials set up, but here is some code that attempts to download all files from an S3 bucket

### DIFF
--- a/datalogistik/util.py
+++ b/datalogistik/util.py
@@ -490,6 +490,21 @@ def decompress(compressed_file_path, output_dir, compression):
         raise ValueError(msg)
 
 
+def download_s3_bucket(url, output_path):
+    import boto3
+    bucket = url.removeprefix("s3://")
+    s3=boto3.client('s3')
+    contents=s3.list_objects(Bucket=bucket)['Contents']
+    for key in contents:
+        object = key['Key']
+        if not object.endswith("/"):
+            s3.download_file(bucket, object, str(output_path) + "/" + object)
+        else:
+            import os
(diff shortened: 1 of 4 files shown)


alistaire47:
> I have no way to test this because I don't have the proper AWS credentials set up

So users will have to have creds to use these sources? What are we doing right now?

datalogistik/util.py
```diff
@@ -490,6 +490,21 @@ def decompress(compressed_file_path, output_dir, compression):
...
+    s3=boto3.client('s3')
+    contents=s3.list_objects(Bucket=bucket)['Contents']
+    for key in contents:
+        object = key['Key']
+        if not object.endswith("/"):
+            s3.download_file(bucket, object, str(output_path) + "/" + object)
+        else:
+            import os
+            if not os.path.exists(object):
+                os.makedirs(object)
```
jonkeane, line 505: Could we use pyarrow here instead of boto3? 

https://arrow.apache.org/docs/python/generated/pyarrow.fs.copy_files.html#pyarrow.fs.copy_files should be able to download all the contents from and s3 url directly with no other dependencies needed

jonkeane (changes requested):
Agreed with @alistaire47 the taxi bucket is public, so we shouldn't need to supply credentials. But generally speaking pyarrow uses "the standard system" credentials since it uses the aws sdk, so if a dataset _is_ private (none of our mvp ones are), setting up the standard way should _just work_

PR #100, opened 2022-09-27
Fix post-download validation

joosthooz:


datalogistik/cli.py
```diff
@@ -97,15 +92,28 @@ def parse_args():
     return parser.parse_args()
 
 
+# Validate all entries in the cache
+def validate_cache(remove_failing):
+    for dir in config.get_cache_location().iterdir():
+        ds = Dataset(name=dir.name)
+        print(ds)
```
jonkeane, line 99: Do we want to keep this in? I don't think the repr method is super nice | if we think folks will use this in python, we should make our own probably...

datalogistik/dataset.py
```diff
@@ -351,39 +432,49 @@ def download(self):
...
+                for file_name in files:
+                    futures.append(
+                        pool.submit(
+                            self.file_listing_item,
+                            os.path.join(cur_path, file_name),
+                        )
+                    )
+            file_list = {}
+            for f in futures:
+                file_list[f.result()["rel_path"]] = f.result()
```
jonkeane, line 466: Would you mind adding a TODO + make an issue in here to measure how much overhead there is to do this with a single file? It might be none, but I suspect there's at least some overhead such that we might want to make a special case for when there's a single file (since that's an overwhelming majority of the cases we have now). We don't need to do that optimization now, but a TODO to remind us to would be great.

datalogistik/dataset.py
```diff
@@ -351,39 +432,49 @@ def download(self):
...
+                table.files = [file_list[str(table_loc.relative_to(path))]]
             elif table_loc.is_dir():
                 # multi file table
                 table.files = [
-                    {
-                        "rel_path": subfile.relative_to(self.ensure_dataset_loc()),
-                        "file_size": subfile.lstat().st_size,
-                        "md5": util.calculate_checksum(subfile),
-                    }
+                    file_list[str(subfile.relative_to(path))]
```
jonkeane, line 477: Do we want `str()` here?

jonkeane (commented):
A few comments (though I see that there are new commits, so they might not be relevant anymore...)

datalogistik/cli.py
joosthooz: This should definitely not have been here, thanks for spotting

datalogistik/dataset.py
joosthooz: Yes, this is just to index into the dict created by the concurrent threads above. The actual value coming out of this is the `file_listing_item`.

datalogistik/dataset.py
joosthooz: I reworked this a bit, also removing some duplication. It took longer than I hoped but I added parallelization back in, only activated for multi-file/table stuff.

datalogistik/dataset.py
```diff
@@ -396,23 +419,19 @@ def download(self):
...
-                    file_name = file.get("file_path")
+                    file_name = file.get("rel_path")
                     if file_name:
                         # All files constituting a table must be in a dir with name table.name (created by ensure_table_loc)
                         download_path = pathlib.Path(file_name).name
                         url = self.url + pathlib.Path(file_name).name
 
                 util.download_file(url, output_path=download_path)
-
-                util.download_file(full_path, output_path=dataset_file_path)
```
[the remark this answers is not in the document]
joosthooz: Oh no worries, pretty sure this is a remnant of some conflict

datalogistik/dataset.py
joosthooz: I changed this code a bit, it's now a helper function creating tuple

datalogistik/dataset.py
```diff
@@ -11,7 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 import concurrent
+import dataclasses
```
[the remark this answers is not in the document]
joosthooz: Yes, done

datalogistik/dataset.py
```diff
@@ -302,6 +302,111 @@ def get_table_dataset(self, table=None):
...
+            "rel_path": rel_path.as_posix(),
+            "file_size": file_size,
+            "md5": file_md5,
+        }
+
+    def create_file_listing(self, table):
+        """Create a file listing for the given table with relative paths, file sizes and md5 checksums."""
+        path = self.ensure_table_loc(table)
+        if path.is_file():
+            return [self.file_listing_item(path)]
```
jonkeane, line 319: A small comment explaining that this is a one-file shortcut would be nice IMO

repo.json
```diff
@@ -12,7 +12,7 @@
                 "header_line": false,
```
jonkeane, line 12: Should we add in the shas, etc. here now that we use them to validate?

jonkeane (commented):
This is getting very close! Just two comments

repo.json
joosthooz: Good point, we definitely should!

Issue #76, opened 2022-09-19
Move `header_line` to Table


jonkeane:
And also ensure that it's default setting makes sense + it is used correctly

jonkeane:
> If a user-specified schema is found, there would usually not be a header line too (but it should be possible - in this case which names would be used?)

Agreed this is a tendency, though it's definitely not consistently always the case like you mention in your parenthetical

alistaire47:
> > If a user-specified schema is found, there would usually not be a header line too (but it should be possible - in this case which names would be used?)
> 
> Agreed this is a tendency, though it's definitely not consistently always the case like you mention in your parenthetical

Yeah, I don't think you can take this for granted, especially when dealing with CSVs and such where you might be using a schema to specify precision or whatever for a dataset written by someone else.



> * Passing a schema is not enough; you also need to separately supply the column names (even though they are also stored in the schema).

In [voltrondata-labs/benchmarks we just grabbed column names from the schema](https://github.com/voltrondata-labs/benchmarks/blob/main/benchmarks/_sources.py#L337); otherwise you have to handle the case where they don't line up

PR #101, opened 2022-09-28
Allow downloading multi file datasets

### LINKED ISSUE #81: Allow me to download multi-file datasets
This is possible, but totally untested (and probably doesn't quite work)

joosthooz:
These modifications are cherry-picked from #76 
Fixes #81 
Currently, this PR doesn't change the behavior of the main branch wrt multi-file downloads; the `url` top-level property is treated as a base, and if there are multiple files in a table, this base is then extended with the filename in the `rel_path` file property.
Note that this will currently cause problems if a dataset has multiple tables and some of them are single- while other multi-file.


datalogistik/dataset.py
```diff
@@ -433,36 +433,43 @@ def download(self):
...
+                    util.set_readonly(download_path)
+
+                # Try validation in case the dataset info contained checksums
+                if not self.validate_table_files(table):
+                    util.clean_cache_dir(dataset_path)
+                    msg = "File integrity check for newly downloaded table failed."
+                    log.error(msg)
+                    raise RuntimeError(msg)
+
+            self.write_metadata()
```
joosthooz, line 468: Apart from the try block, this was added (so that the calls to the metadata functions in `datalogistik/dataset_search.py` could be removed)

joosthooz:
So the question remains whether to use the base url + suffixes per file, or to have a full url per file. I have a preference for the latter. If we stick with the first, we should still probably move the base url property to be a per-table item, so that we can handle cases where some tables are single-file and others are multi-file.

datalogistik/dataset.py
jonkeane: That makes sense — do we also need `dataset.fill_metadata_from_files()` here as well?

joosthooz: Not anymore, that was added into `write_metadata()`

joosthooz:
Here's the approach I took in the last set of commits (from the comments in the code):
# There are 3 possible types downloads:
# 1 - single table, single file. url can be a top-level property, just download it.
# 2 - multi table, single file. The table entry has a url property.
# 3 - multi file (either single or multi table). The table entry has a base_url property,
# and each file has a rel_path property. This is appended to the base_url to form
# the download link. The files will be placed in the table directory (generated from the table name).

joosthooz:
> We can collapse 1 + 2 together if we always put the one-true-URL into the table entry, yeah?
Yes, and I think that this is a better way to do it. Actually it seems that most of the properties should move down to the table entry, for example also `delim` and `compression`! Maybe something for a next PR though,
> How do I specify 1 table, multi file, but with 1 url?
This will not work right now, but it is also not supported by the current download code. We need to add S3 support for this. I already have the code for that in a branch, the question is do we want to include it in this PR?
> mock the actual function with something that moves a (set of, small) fixture file(s)
This is a nice idea and I think it should be doable. Right now there's some mock functions that just check the values that would be passed to the download function, and this does allow testing the validation step we added in #100.

jonkeane:
> > We can collapse 1 + 2 together if we always put the one-true-URL into the table entry, yeah?

> Yes, and I think that this is a better way to do it. Actually it seems that most of the properties should move down to the table entry, for example also delim and compression! Maybe something for a next PR though,

I'm wary of moving too many things down into the table entry. But most specifically: we should not (without very very very great evidence this is critical) move anything down to the table entity that is or reasonably will be a setting someone might tune (like compression) since those things should be applied to each table collectively (at least after the initial raw -> transformed state, we might need to support a set of tables that have different raw specifications, but let's deal with that in the future if it ever truly comes up).

> How do I specify 1 table, multi file, but with 1 url?

> This will not work right now, but it is also not supported by the current download code. We need to add S3 support for this. I already have the code for that in a branch, the question is do we want to include it in this PR?

We don't need to add the S3 code to support this, but if we're going to make rules about the structure of our repo.json, we should not make rules that would make this impossible + hard + yet another case in a list of possible configurations here.

joosthooz:
Ok, I think I addressed most of the active discussions:
- I removed support for a top-level `url` property, but kept in in a branch in case we want to try it out. To use it, there should not be a `tables` entry. This allows very simple entries in the repo file (just a name and url). There's a simple detection for format and compression.
- There's now 2 more tests for downloads, that both include a mock download function that just copies a file to a tmp dir. 1 has a failing checksum, the other succeeds.
- Multi-file downloads are now supported in the way we discussed. We only need to add support for it in the download function.
- I moved all the urls in the repo file down to the table entries.

tests/test_dataset.py
```diff
@@ -256,44 +258,83 @@ def _fake_download(url, output_path):
...
-            simple_parquet_ds.ensure_dataset_loc(),
-            tmpcachepath + "/" + simple_parquet_ds.name,
+    with tempfile.TemporaryDirectory() as tmpcachedir:
+        tmpcachepath = pathlib.Path(tmpcachedir)
+
+        def _simulate_download_by_copying(url, output_path):
+            shutil.copyfile(simple_parquet_ds.ensure_table_loc(), output_path)
+
+        monkeypatch.setattr(
+            "datalogistik.util.download_file", _simulate_download_by_copying
```
austin3dickey, line 278: Nice

austin3dickey (approved):
Looking good

datalogistik/dataset.py
```diff
@@ -433,43 +444,87 @@ def download(self):
...
+                        msg = f"Single-file table '{table.table}' has 'base_url' property set. It should only have a 'url'."
+                        log.error(msg)
+                        raise ValueError(msg)
+                    for file in table.files:
+                        # contains the suffix for the download url
+                        rel_path = file.get("rel_path")
+                        if not rel_path:
+                            msg = f"Missing rel_path property for multi-file table '{table.table}'."
+                            log.error(msg)
+                            raise ValueError(msg)
```
jonkeane, line 493: What do you think about using `url` for both types, and then treating it as a base URL IFF the files have `rel_paths` associated with them?

joosthooz: I thought about it but prefer distinguishing the 2, because the `rel_path` is also used for checksum verification and will often be present for single-file tables too. So then it would force the use to have to split the url up which feels strange. But it does prevent weirdness that would happen if those don't match (the filename in the url and the rel_name in the file entry - in this case they are not the same the validation would just fail).

jonkeane: Hmm, that _also_ sounds like a problem (which could crop up if we do or don't resolve the url|base_url issues...) It sounds like we are using `rel_path` to refer to both a relative path in the cache as well as a relative URL path for downloading. Is there any way that we could differentiate those uses?

PR #106, opened 2022-10-10
56 remote dataset support

### LINKED ISSUE #56: Remote dataset support
[In arrowbench](https://github.com/voltrondata-labs/arrowbench/blob/main/R/bm-remote-dataset.R) we test querying remote datasets, in which case the data source merely provides the metadata for what bucket or whatever it is in, and doesn't download anything. Do we want to support this use-case? What would the interface look like?

joosthooz:
Using #101 as target branch for now, because these changes depend on it.
Fixes #56, adds support for s3.
Adds a new property to repo file entries; `remote`. These entries cannot be downloaded or converted. The output function will simply passthrough the url.


tests/test_dataset.py
```diff
@@ -482,6 +482,25 @@ def test_output_result():
...
+            "format": "parquet",
+            "tables": {
+                "remote_table": {
+                    "path": "s3://remote_table/",
+                    "dim": [],
+                },
+            },
+        }
+    )
+    assert Dataset(
```
austin3dickey, line 497: Did you mean to compare this to the `expected` variable?

joosthooz: Absolutely, thanks for spotting this

datalogistik/util.py
```diff
@@ -374,11 +375,14 @@ def download_file(url, output_path):
         shutil.rmtree(output_path, ignore_errors=True)
 
     try:
-        http = urllib3.PoolManager()
-        with http.request("GET", url, preload_content=False) as r, open(
-            output_path, "wb"
-        ) as out_file:
-            shutil.copyfileobj(r, out_file)  # Performs a chunked copy
+        if url[0:2] == "s3":
+            pyarrow.fs.copy_files(url, output_path)
```
jonkeane, line 379: Would you mind adding docs somewhere that explains that the s3 URL being downloaded will be downloaded recursively? So if you have a directory of a bunch of files, you'll get the whole thing?

austin3dickey:
>any dataset in the repo.json could be used as a remote dataset

I think this makes sense. So this would look like a `--remote` flag in the CLI that returns the "remote" version of the dataset?

jonkeane:
> I think this makes sense. So this would look like a --remote flag in the CLI that returns the "remote" version of the dataset?

Yeah, that sounds good

joosthooz:
I added a CLI option. But I kept `remote` as a field in `Dataset`, so technically it is still possible to specify a dataset in the repo file as "remote-only". It will return the remote url regardless of whether the CLI option was used.

jonkeane:
> I added a CLI option. But I kept remote as a field in Dataset, so technically it is still possible to specify a dataset in the repo file as "remote-only". It will return the remote url regardless of whether the CLI option was used.

IMO that sounds pretty confusing. The benchmark runners (the main 'audience' for datalogistik) know better if they should or should not be using a remote source or not — seems off to me to override that choice based on the repo

austin3dickey:
>The benchmark runners know better

+1. Even if the repo doesn't override the user choice, I think in all cases the user will know whether they need a remote dataset or not. Cutting it from the repo file will help cut down our future maintenance of the repo file and its associated code too.

joosthooz:
The thing is, in that case I need a different way of communicating that option from CLI parsing back to main(). Because then it cannot be a Dataset field (because then it would be something that could be specified in the repo entries). Any thoughts?

austin3dickey:
I see what you're saying, since we only pass a Dataset instance between the CLI and `main()`, and the `output_result()` code is a Dataset method anyway, it would be nice to store `remote` as an attribute on the Dataset instance we're passing around in memory. 

Could we make that a hidden attribute; e.g. `_remote`? So that it isn't stored in the repo. And ensure that:
- `write_metadata()` and `to_json()` skip hidden attributes (this might already be okay with the default dataclasses behavior, not sure)
- hidden attributes aren't checked for equality between two Datasets

joosthooz:
It seems to be not so easy to create a hidden field using dataclasses (see https://florimond.dev/en/posts/2018/10/reconciling-dataclasses-and-properties-in-python/), but I renamed it to `_remote` which is explicitly removed in `to_json()` and produces an error if present in the repo file.

joosthooz:
Merging after some pair-programming with @jonkeane

PR #107, opened 2022-10-10
53 feather file conversion support

### LINKED ISSUE #53: Feather file conversion support
We need to be able to save files as feather files, uncompressed, lz4, and zstd  compression for [arrowbench](https://github.com/voltrondata-labs/arrowbench/blob/main/R/bm-read-file.R#L13) and [voltrondata-labs/benchmarks](https://github.com/voltrondata-labs/benchmarks/blob/main/benchmarks/file_benchmark.py#L15-L18).

joosthooz:
Fixes #53 
Adds support for the feather file support, although conversion to compressed feather does not work because this is not supported by pyarrow.dataset. In addition, there is no API to find the compression used in an existing file (like for parquet), so we rely on the user to set this property in the metadata.

This PR also adds a parametrized test for converting between the 3 supported formats, with randomly generated data of a schema that includes all types that are supported by all 3 formats.


datalogistik/dataset.py
```diff
@@ -529,6 +544,11 @@ def convert(self, new_dataset):
             f"compression {self.compression} to {new_dataset.format}, "
             f"compression {new_dataset.compression}..."
         )
+        if new_dataset.format == "feather" and new_dataset.compression:
+            msg = "Compression not supported when converting to Feather format."
+            log.error(msg)
+            raise ValueError(msg)
+
```
jonkeane, line 551: We might need to upstream adding this option to pyarrow — we will need this compression available in order to use datalogistik. This _is_ available from R (cf https://arrow.apache.org/docs/r/reference/write_dataset.html `write_dataset(..., codec = Codec$create("lz4"))` works for example.

https://github.com/apache/arrow/blob/3d7f2f22a0fc441a41b8fa971e11c0f4290ebb24/r/src/dataset.cpp#L257-L264 is the cpp that wires it up

tests/test_datalogistik.py
```diff
@@ -296,10 +370,53 @@ def test_compress(comp_string):
...
+@pytest.mark.parametrize("dest_format", ["csv", "parquet", "feather"])
+def test_convert_parquet(monkeypatch, source_format, dest_format):
+    if source_format == dest_format:
+        pytest.skip()
+    name = "data_to_be_converted"
+    file_name = name + "." + source_format
+    data = generate_complete_schema_data(100, "common")
+    orig_table = pa.table(data, schema=common_schema)
+    with tempfile.TemporaryDirectory() as tmpdspath:
+        monkeypatch.setenv("DATALOGISTIK_CACHE", tmpdspath)
```
jonkeane, line 383: Could we set this at the top of the file like we do at https://github.com/conbench/datalogistik/blob/be1f24c99e8d78ac4af9d3490099a3a5e54679b4/tests/test_dataset.py#L111-L113 ?

joosthooz:
Created JIRA about the absence of compression support in pyarrow.dataset.ipc: https://issues.apache.org/jira/browse/ARROW-17991

README.rst
```diff
@@ -32,7 +32,7 @@ Usage::
 ``FORMAT``
     File format to instantiate the dataset in. If the original dataset specified in the
     repository has a different format, it will be converted. Supported formats:
-    ``parquet``, ``csv``.
+    ``parquet``, ``csv``, ``feather``.
```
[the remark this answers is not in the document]
joosthooz: We need Arrow 10.0 for that (https://github.com/apache/arrow/pull/13677).
I pushed the changes needed in datalogistik here https://github.com/conbench/datalogistik/pull/109

tests/test_datalogistik.py
joosthooz: We cannot, because the tempfile only exists in this context. Is this a bad pattern? Should I do a cleanup like here https://github.com/conbench/datalogistik/blob/be1f24c99e8d78ac4af9d3490099a3a5e54679b4/tests/test_datalogistik.py#L338 ?

austin3dickey: I think the way it is now is fine. It's something that won't be re-used across different tests

joosthooz:
At first, I figured we'd just use the normal ipc writer for now, but now you mention the nightlies, I think that might be a better idea. Because there could be more stuff we need that we add to Arrow ourselves.

jonkeane:
Would you mind setting up the dependency setup to pull a nightly for arrow in this PR + take out the stops that prevent compression from being used?

joosthooz:
Done. also merged #109 into this, so will close it.

joosthooz:
Now, pyarrow nightlies are installed for the test-wheel step. But for pipx, I had to add the flags (most notably `--pre`) needed to install pyarrow nightlies to the pipx command to install datalogistik, because otherwise a separate environment would be created. But I guess that means that for the other dependencies it might also install non-stable packages.

.github/workflows/actions.yml
```diff
@@ -120,8 +120,15 @@ jobs:
...
-        run: ${{ matrix.installer }} install '${{ matrix.installee }}'
+      - name: Pip install package
+        if: matrix.installer == 'pip'
+        run: |
+          ${{ matrix.installer }} install --extra-index-url https://pypi.fury.io/arrow-nightlies/ --prefer-binary --pre pyarrow
+          ${{ matrix.installer }} install '${{ matrix.installee }}'
+      - name: Pipx install package
+        if: matrix.installer == 'pipx'
+        run: |
+          ${{ matrix.installer }} install --pip-args "--extra-index-url https://pypi.fury.io/arrow-nightlies/ --prefer-binary --pre" '${{ matrix.installee }}'
```
jonkeane, line 131:
```suggestion
          ${{ matrix.installer }} install --pip-args "--extra-index-url https://pypi.fury.io/arrow-nightlies/ --prefer-binary --pre=pyarrow" '${{ matrix.installee }}'
```

Does this work to restrict pyarrow to be the only prerelease allowed?

datalogistik/dataset.py
```diff
@@ -682,8 +699,7 @@ def convert(self, new_dataset):
             log.debug(f"conversion took {conv_time:0.2f} s")
 
             new_dataset.write_metadata()
-
-            util.set_readonly_recurse(output_file)
```
jonkeane, line 686: I'm surprised this line wasn't erroring before?

datalogistik/dataset_search.py
```diff
@@ -70,8 +70,9 @@ def find_or_instantiate_close_dataset(dataset):
     # when we support .arrow, those likely should be first, then parquet, et c.
```
jonkeane, line 70: We should rephrase this comment now that we support arrow files

tests/test_datalogistik.py
```diff
@@ -98,14 +104,14 @@ def test_version():
...
-    "x": {"type_name": "timestamp", "arguments": {"unit": "ms"}}
+    "x": {"type_name": "timestamp", "arguments": {"unit": "us"}}
     }"""
 complete_csv_schema_json_output = (
     "{'a': 'null', 'b': 'bool', 'c': 'int8', 'd': 'int16', 'e': 'int32', 'f': 'int64', "
     "'g': 'uint8', 'h': 'uint16', 'i': 'uint32', 'j': 'uint64', 'l': 'float', 'm': "
     "'double', 'n': 'date32[day]', 'o': 'date64[ms]', 'q': 'string', 'r': 'string', "
     "'t': 'large_string', 'u': 'large_string', 'v': 'time32[ms]', 'w': 'time64[us]', "
-    "'x': 'timestamp[ms]'}"
+    "'x': 'timestamp[us]'}"
```
jonkeane, line 114: Can you explain more about this change? It seems a bit unrelated, was it needed for the tests to pass or something?

tests/test_datalogistik.py
```diff
@@ -133,13 +139,47 @@ def test_version():
...
+#         pa.field("u", pa.large_utf8()),
+#         # types with arguments
+#         pa.field("v", pa.time32("ms")),
+#         pa.field("w", pa.time64("us")),
+#         pa.field("x", pa.timestamp("us")),
+#         # pa.field("y", pa.duration("s")), # not supported by parquet and csv
+#         pa.field("z", pa.binary(10)),
+#         pa.field("argh", pa.decimal128(7, 3)),
+#     ]
+# )
```
jonkeane, line 179: Can we remove this commented code?

jonkeane (changes requested):
Looking good, a few small(ish) changes we need to make, but once those are made  I tink we're ready to merge + try this in arrowbench

tests/test_datalogistik.py
joosthooz: IIRC, one of the formats (can't remember which) didn't support ms, or otherwise had a problem with round-tripping it to the other formats.

tests/test_datalogistik.py
joosthooz: Done

datalogistik/dataset_search.py
joosthooz: Done

datalogistik/dataset.py
```diff
@@ -297,11 +297,11 @@ def get_table_dataset(self, table=None):
...
         table = self.get_one_table(table)
 
-        # Defaults
-        schema = table.schema
+        schema = None
         if self.format == "parquet":
             dataset_read_format = pads.ParquetFileFormat()
-            schema = None
+        if self.format == "arrow":
+            dataset_read_format = "arrow"
```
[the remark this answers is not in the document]
joosthooz: Absolutely!

.github/workflows/actions.yml
joosthooz: I tried a different approach; instead of the general `--pre` (removed), I added a dev version as minimum (`'pyarrow>=10.0.0dev497'`).

joosthooz: After a few tries, it finally works, by just adding the nightly repo to the install commands and then the new minimum version sorts it out automatically.

.github/workflows/actions.yml
```diff
@@ -120,8 +120,14 @@ jobs:
...
-      - name: Install package
-        run: ${{ matrix.installer }} install '${{ matrix.installee }}'
+      - name: Pip install package
+        if: matrix.installer == 'pip'
+        run: |
+          ${{ matrix.installer }} install --extra-index-url https://pypi.fury.io/arrow-nightlies/ --prefer-binary '${{ matrix.installee }}'
+      - name: Pipx install package
+        if: matrix.installer == 'pipx'
+        run: |
+          ${{ matrix.installer }} install --pip-args "--extra-index-url https://pypi.fury.io/arrow-nightlies/ --prefer-binary" '${{ matrix.installee }}'
```
jonkeane, line 130: We should add the extra index urls to the readme too

jonkeane (approved):
I tested this with arrowbench and it works 🎉 . One comment about adding the pipx instructions to the readme (and looks like there's a conflict too)

PR #108, opened 2022-10-10
Added option to add generator path using env var

### LINKED ISSUE #68: Should the `generator-path` argument actually be an env var?
This seems like something someone would want to set + forget and not need to pass through each time.

joosthooz:
Fixes #68 
Adds the option to add generator path using env var `DATALOGISTIK_GEN`.


austin3dickey (approved):
lgtm!

PR #109, opened 2022-10-12
Change file extension from .feather to .arrow

joosthooz:
When Arrow 10.0 is release (we need this https://github.com/apache/arrow/pull/13677), we can use these changes to update the file extension of Arrow IPC files to .arrow.


austin3dickey:
Just to clarify, we won't merge this until Arrow 10 is released?

joosthooz:
> Just to clarify, we won't merge this until Arrow 10 is released?

Affirmative! Also, I just changed the target branch so it's more clear that this is just a name change after merging #107

Issue #111, opened 2022-10-17
Differentiate rel_path as url suffix and relative path in cache


joosthooz:
we are using `rel_path` to refer to both a relative path in the cache as well as a relative URL path for downloading. Is there any way that we could differentiate those uses?

_Originally posted by @jonkeane in https://github.com/conbench/datalogistik/pull/101#discussion_r997052557_

Note that this might just be a documentation issue; my first intuition is that we should indeed only support cases in which these 2 are the same. rel_path is the path to the file relative to the table's directory.

joosthooz:
Resolved by https://github.com/conbench/datalogistik/pull/113

PR #113, opened 2022-10-20
Separate `rel_path` from `rel_url_path`, use one `url`

### LINKED ISSUE #110: Possibly unify url and base_url
What do you think about using `url` for both types, and then treating it as a base URL IFF the files have `rel_paths` associated with them?

_Originally posted by @jonkeane in https://github.com/conbench/datalogistik/pull/101#discussion_r994785725_

Keep in mind that in that case, `rel_path` and the filename in the url must be identical, otherwise validation will fail.

jonkeane:
Resolves #110 and #111

This introduces a disambiguation of `rel_path` from `rel_url_path`. The idea is that:

* `rel_path` is that path _in the cache, on disk_
* `rel_url_path` is the path needed for downloading only

This distinction is really only needed for cases where the table name (which is what a single-file table will be named once downloaded) is different from the path that it is stored as. We might not even need this "feature" though we have [a table where that happens | is true already](https://github.com/conbench/datalogistik/blob/c41e6426364211a556090a0d1e070739858fc0e4/repo.json#L71-L72). We could also overwrite | change `rel_path` after we download the file, but that doesn't feel great (in case we maybe want to redownload? maybe it's not worth maintaining at all...)

`rel_url_path` is really only needed in cases where there is a multi-file table, where the files are listed (e.g. not the case of the taxi s3 bucket, we don't yet have examples of this!). But it can be specified elsewhere if that's helpful.

All of this means that: for the easy path, an author of a new repo entry generally just uses `url` and that's it. But we can save typing if there are multi-files by using `url` as a base with `rel_url_paths` added on. And then we (internally! with code) maintain the `rel_path` `rel_url_path` if we need those.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -273,13 +273,13 @@ datalogistik_metadata.ini
         A list of files in this table. Each entry in the list has the following properties:
 
         ``rel_path``
-            Path to the file(s), relative to the directory of this table. This is the the 
+            Path to the file(s), relative to the directory of this table. This is the 
             location on disk in the cache.
         
         ``rel_url_path``
-            URL path to the file(s), relative to the directory of this table. This is used 
-            only when downloading the file. This is only necesary when a multi table file has
-            the files that make up the table listed out individually. 
+            URL path to the file(s), relative to the directory of this table where it is stored 
+            remotely. This is used only when downloading the file. This is only necesary when a 
+            multi table file has the files that make up the table listed out individually. 
 
         ``file_size``
             Size of the file.
--- a/README.rst
+++ b/README.rst
@@ -263,17 +263,23 @@ datalogistik_metadata.ini
         Schema of the table.
 
     ``url``
-        Download url in case this is a single-file table.
-
-    ``base_url``
-        Base download url in case this is a multi-file table. Each file will append
-        their `rel_path` to this to form the full download url.
+        Download url for the table. This can be: 
+        * A URL specifying the file to be downloaded for that table (which could be a 
+          single file, or a directory that contains many files to be downloaded)
+        * A base URL that is concatenated with ``rel_url_path``s in the ``files`` attribute 
+          if the table is a multi-file table and it is preferable to list out the files
 
     ``files``
         A list of files in this table. Each entry in the list has the following properties:
 
         ``rel_path``
-            Path to the file, relative to the directory of this table.
+            Path to the file(s), relative to the directory of this table. This is the the 
+            location on disk in the cache.
+        
+        ``rel_url_path``
+            URL path to the file(s), relative to the directory of this table. This is used 
+            only when downloading the file. This is only necesary when a multi table file has
+            the files that make up the table listed out individually. 
 
         ``file_size``
             Size of the file.
--- a/datalogistik/dataset.py
+++ b/datalogistik/dataset.py
@@ -357,7 +357,6 @@ class Dataset:
                 f"No metadata found for table {table}, could not perform validation (assuming valid)"
             )
             return True
-
         new_file_listing = self.create_file_listing(table)
         # we can't perform a simple equality check on the whole listing,
         # because the orig_file_listing does not contain the metadata file.
@@ -372,6 +371,8 @@ class Dataset:
             found = None
             for new_file in new_file_listing:
                 if new_file["rel_path"] == orig_file["rel_path"]:
+                    # drop the rel_url_path for comparison, because it's not relevant!
+                    orig_file.pop("rel_url_path", None)
                     found = new_file
                     break
 
@@ -448,7 +449,7 @@ class Dataset:
             msg = (
                 "No table entries were found. "
                 "To download a dataset, at least 1 table entry must exist "
-                "that has a 'url' or 'base_url' property."
+                "that has a 'url' property."
             )
             log.error(msg)
             raise ValueError(msg)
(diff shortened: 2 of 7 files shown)


datalogistik/dataset.py
```diff
@@ -372,6 +371,8 @@ def validate_table_files(self, table):
             found = None
             for new_file in new_file_listing:
                 if new_file["rel_path"] == orig_file["rel_path"]:
+                    # drop the rel_url_path for comparison, because it's not relevant!
+                    orig_file.pop("rel_url_path", None)
```
jonkeane, line 375: We might want to do the reverse of this some day and only check a subset of specific fields instead of dropping this, but this was the easiest way to working code.

tests/test_dataset.py
```diff
@@ -249,17 +250,71 @@ def test_validate():
...
+            "chi_traffic_2020_Q1.parquet",
+            "https://ursa-qa.s3.amazonaws.com/chitraffic/chi_traffic_2020_Q1.parquet",
+        ),
+        # we can have a tablename that is differnet from rel_url_path
+        (
+            "https://ursa-qa.s3.amazonaws.com/chitraffic",
+            "traffic.parquet",
+            "https://ursa-qa.s3.amazonaws.com/chitraffic/traffic.parquet",
+        ),
+    ],
```
jonkeane, line 279: This is probably overkill (basically to support someone specifying the full url _and_ the rel_path), but it's a handful of lines of code + some tests to cover a relatively confusing bit. We can pull it out (or turn the fix-up code that removes the filename if it's at the end of the url into validation that tells someone they've messed up their repo code)

tests/test_dataset.py
```diff
@@ -249,17 +250,71 @@ def test_validate():
...
+    def _fake_write_metadata(self):
+        return True
+
+    monkeypatch.setattr(Dataset, "write_metadata", _fake_write_metadata)
+
+    # This probably also means that set_readonly isn't in the right place, but let's mock it away nonetheless
+    def _fake_set_readonly(path):
+        return True
+
+    monkeypatch.setattr("datalogistik.util.set_readonly", _fake_set_readonly)
```
jonkeane, line 301: In working on these tests I've left things in place as much as possible, but that we need to monkey patch these functions tells me we probably don't have `Dataset.download()` factored like we should. I would propose we have a method that does the downloading and no validation or anything (maybe Dataset._download()`), and then we wrap that with the validation etc. stuff later (so then when we test that we only need to mock the actual downloading bit and not also permissions+metadata writing+permissions munging). 

https://github.com/conbench/datalogistik/issues/114

tests/test_dataset.py
```diff
@@ -268,7 +323,46 @@ def _fake_download(url, output_path):
...
+
+    # add rel_url_path as if this were coming from a repo file with that.
+    ds = copy.deepcopy(multi_file_ds)
+    files = ds.tables[0].files
+    # add rel_url_path
+    [fl.update({"rel_url_path": fl["rel_path"]}) for fl in files]
+    # remote rel_path
+    [fl.pop("rel_path") for fl in files]
```
jonkeane, line 358: This is a bit funny, but it's because we're using metadata from the fixture cache to then test downloading. So we need to add `rel_url_path` in here as if it were in `repo.json`

We could also add `rel_url_path` in to the fixtures and take out some of this (though it's good that we're testing the `rel_path` setting which (rightfully) happens during downloading.

tests/test_dataset.py
```diff
@@ -298,7 +392,8 @@ def _simulate_download_by_copying(url, output_path):
             # Note that if we were not working in a tmp dir, the dir should have been deleted
             # and the exception message would be: "File integrity check for newly downloaded dataset failed."
 
-    # Test a succeeding download
+
+def test_validated_download_dataset(monkeypatch):
```
jonkeane, line 396: I broke these out to make them a bit easier to see which failed when. Happy to put them back if we're 👎 on that

datalogistik/dataset.py
```diff
@@ -460,51 +461,48 @@ def download(self):
...
+
+                    # if the filename is not at the end of full_path, join
+                    table_url = table.url
+                    if rel_url_path and not table_url.endswith(rel_url_path):
+                        table_url = table_url + "/" + rel_url_path
+
+                    download_path = table_path
+
+                    # Set the rel_path
+                    file["rel_path"] = download_path.name
```
joosthooz, line 487: Why overwrite `file["rel_path"]` here? Don't we need to it figure out where to place the file? The contents could be different from `download_path.name`, especially if it is placed in a subdir (because of partitioning for example), and then the validation would fail

Or did I misunderstand and is `rel_path` now not a user input anymore? It seems I did, but then my question is; if `rel_url_path` is set, it is alway identical to `rel_path`, so do we need the distinction?

tests/test_datalogistik.py
```diff
@@ -299,7 +299,10 @@ def test_compress(comp_string):
 # Integration-style tests
 def test_main(capsys):
     # This should be in the cache already, so no conversion needed
-    exact_dataset = dataset.Dataset(name="fanniemae_sample", format="csv", delim="|")
+    # import pdb; pdb.set_trace()
+    exact_dataset = dataset.Dataset(
+        name="fanniemae_sample", format="csv", delim="|", compression="gzip"
```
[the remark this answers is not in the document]
joosthooz: Oh wow good catch, it has proved rather difficult to get the default behavior right in all situations! Maybe we should reconsider the `None` versus "uncompressed" distinction (where `None` is basically "dont-care"), then our equality operator could just return equal if one of the objects has compression "dont-care"

tests/test_dataset.py
joosthooz: I like it this way, because it's nice to still support a simple url (that is not made up by a base + rel_url_path)

tests/test_dataset.py
joosthooz: We could add `rel_url_path` in the metadata files in the fixture cache, I think these properties will be carried from the repo file entry so they would be there in a real metadata file too

README.rst
```diff
@@ -263,17 +263,23 @@ datalogistik_metadata.ini
...
-        Download url in case this is a single-file table.
-
-    ``base_url``
-        Base download url in case this is a multi-file table. Each file will append
-        their `rel_path` to this to form the full download url.
+        Download url for the table. This can be: 
+        * A URL specifying the file to be downloaded for that table (which could be a 
+          single file, or a directory that contains many files to be downloaded)
+        * A base URL that is concatenated with ``rel_url_path``s in the ``files`` attribute 
+          if the table is a multi-file table and it is preferable to list out the files
```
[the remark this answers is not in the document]
austin3dickey: I think you just need more line breaks (usually the solution to RST woes :) ). Also watch out for the "``s"
```
    ``url``

        Download url for the table. This can be:

        * A URL specifying the file to be downloaded for that table (which could be a 
          single file, or a directory that contains many files to be downloaded)
        * A base URL that is concatenated with ``rel_url_path`` in the ``files`` attribute 
          if the table is a multi-file table and it is preferable to list out the files
```

tests/test_dataset.py
austin3dickey: Nice

datalogistik/dataset.py
jonkeane: > it is alway identical to rel_path, so do we need the distinction?

You're right that it's (almost) always identical, but there are two cases where it's not identical:
* in the multi-file table scenario below
* when the table name is different from the filename

PR #116, opened 2022-11-01
Added ndjson support

### LINKED ISSUE #55: ndjson file conversion support
[We currently benchmark ndjson file reading](https://github.com/voltrondata-labs/arrowbench/blob/main/R/bm-read-json.R), which requires converting a dataframe to an ndjson file, which can be either uncompressed (`.json`) or gzipped (`.json.gz`). Nulls should be explicit, i.e. `..., "col3": null, ...`.

joosthooz:
Resolves #55 
Adds support for ndjson. Both convert from and to ndjson works. For reading, pyarrow.json is used (note: pyarrow.dataset does not support ndjson, so we have to read a dataset fully into memory). For writing, the ndjson module is used.


joosthooz:
This is WIP; there seems to be something wrong when converting _from_ ndjson to the other formats.

joosthooz:
Now it works. But I still want to make some improvements to the conversion tests.

joosthooz:
I made some improvements that should make the test code a bit more clear and remove a lot of duplication from the schema (all kinds of variants for the different format, each with a subset of suppoted types).

datalogistik/dataset.py
```diff
@@ -634,65 +638,95 @@ def convert(self, new_dataset):
...
+                        )
+                    else:
+                        table = table_pads.to_table()
+                    nrows = table.num_rows
+                    ncols = len(table.schema.names)
+
+                    with open(output_file, "w") as f:
+                        writer = ndjson.writer(f)
+                        for row in table.to_pylist():
+                            writer.writerow(row)
```
jonkeane, line 665: I wonder if this is getting complicated enough that it makes sense to pull it out into a new function?

datalogistik/dataset.py
```diff
@@ -310,6 +312,8 @@ def get_table_dataset(self, table=None):
             dataset_read_format, schema = self.get_csv_dataset_spec(table)
         if self.format == "tpc-raw":
             dataset_read_format, schema = self.get_raw_tpc_dataset_spec(table)
+        if self.format == "ndjson":  # not supported by pyarrow.dataset
```
jonkeane, line 315: It might be nice to expand on this comment a little bit too, explaining that we use a table + will pull into memory, etc.

tests/test_datalogistik.py
```diff
@@ -338,25 +293,56 @@ def test_compress(comp_string):
...
+    subset_json_schema = complete_schema_json_input
+    unsupported_columns = []
+    if "csv" in [source_format, dest_format]:
+        unsupported_columns.extend(["k", "p", "s", "y", "z", "argh"])
+    if "parquet" in [source_format, dest_format]:
+        unsupported_columns.extend(["i", "k", "o", "p", "y"])
+    if "ndjson" in [source_format, dest_format]:
+        unsupported_columns.extend(
+            ["k", "n", "o", "p", "s", "v", "w", "x", "y", "z", "argh"]
+        )
```
jonkeane, line 307: While we're in here, would it be possible to name the columns with monikers that are close to the types that they are, so it's easier to read this test as "ah, if it's a csv I don't support types: int32, float128, etc."?

jonkeane (changes requested):
Looking good. I made two comments where will likely be pretty big changes (at least lines-changed wise). The factoring into a function our writting code is something I was already close to suggesting when we added the arrow format, but looking at how many levels of indentation we have there: I suspect now is the time for us to factor that out into separate functions

jonkeane:
I haven't had a chance to dig into this more deeply, but I'm getting a funny error on linux only: https://github.com/voltrondata-labs/arrowbench/actions/runs/3395986451/jobs/5646533349#step:10:272

joosthooz:
It seems the memory usage was too high, so the program would get killed when convering a large dataset like fanniemae. This last change should help.

datalogistik/dataset.py
jonkeane: Thoughts about ^^^?

tests/test_datalogistik.py
jonkeane: I still think that ^^^ would be a good idea — is that too much work for this PR?

jonkeane (changes requested):

datalogistik/dataset.py
jonkeane: Thoughts on this? I think factoring this stuff out into separate functions would really improve readability + testability

jonkeane:
I just rebased this to get the new env var for not-changing permissions, testing over an arrowbench again

datalogistik/dataset.py
joosthooz: Absolutely, I gave it a try. There is some uglyness because we need to pass the source format and file path in case we're dealing with ndjson.
Also, the writing function is a member function, which should be called on the `new_dataset`, which could a bit confusing. Maybe I should pass a complete source Dataset instead. What do you think?

datalogistik/dataset.py
joosthooz: Definitely good to expand this comment a bit, done

tests/test_datalogistik.py
joosthooz: I converted the names to basically the type name, I think this makes it much more clear!

PR #117, opened 2022-11-02
Clarifying Dataset's cache_location field

### LINKED ISSUE #80: Clarify the path locations + how to construct those paths
And also clean up the names of attributes (e.g. `Dataset.cache_location` is the location _inside of_ the cache, but that's a bit confusing for that name!)

Right now: the path attribute in the dataset is (or should be if it is not) relative to the cache location and the path attribute in the (files list of the) table attribute is relative to the dataset. We should enforce this + document this better than it is. That makes constructing paths to be always something like config.cache_location / dataset.ensure_dataset_loc() / dataset.ensure_table_loc()

joosthooz:
resolves #80 
Changed cache_location to full_path, and full_path is now set based on the location of the given metadata file in from_json().

### DIFF
--- a/datalogistik/dataset.py
+++ b/datalogistik/dataset.py
@@ -62,13 +62,12 @@ class Dataset:
     local_creation_date: Optional[str] = None
     scale_factor: Optional[float] = None
     delim: Optional[str] = None
-    metadata_file: Optional[pathlib.Path] = None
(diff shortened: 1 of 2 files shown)


jonkeane (approved):
Looks good — thanks for this!

PR #118, opened 2022-11-02
Added type hints + docstrings to methods

joosthooz:
Resolves #79 
This also removes some intermediate timers for conversion, generation, and downloading, but it leaves the end2end timer in place.

PR #121, opened 2022-11-08
Let me prevent permissions changes

jonkeane:
This will prevent various CI errors I'm seeing with arrowbench, since we create a cache in a temp dir, but that temp dir is not able to be cleaned up. https://github.com/voltrondata-labs/arrowbench/actions/runs/3421285685/jobs/5697219852#step:11:74

### DIFF
--- a/datalogistik/generate_dataset.py
+++ b/datalogistik/generate_dataset.py
@@ -1,0 +1,74 @@
+# Copyright (c) 2022, Voltron Data.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import time
+
+from . import config, tpc_info
+from .log import log
+from .table import Table
+from .tpc_builders import DBGen, DSDGen
+from .util import clean_cache_dir
+
+
+# Generate a dataset by calling one of the supported external generators
+# TODO: Generator output cannot be used as dataset output directly, because of the
+# trailing columns.
+def generate_dataset(dataset):
+    log.info(f"Generating {dataset.name} data to cache...")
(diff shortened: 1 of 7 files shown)


datalogistik/util.py
```diff
@@ -18,32 +18,39 @@
...
-from .tpc_builders import DBGen, DSDGen
+
+
+def getenv_bool(name: str, default: bool = False) -> bool:
+    return os.getenv(name, str(default)).lower() in ("true", "t", "yes", "y", "1")
 
 
 # Set file permissions of given path to readonly
 def set_readonly(path):
+    if getenv_bool("DATALOGISTIK_NO_PERMISSIONS_CHANGE"):
```
austin3dickey, line 38: Could you add this to README?

austin3dickey (approved):
Nice

datalogistik/generate_dataset.py
```diff
@@ -0,0 +1,74 @@
...
+from .log import log
+from .table import Table
+from .tpc_builders import DBGen, DSDGen
+from .util import clean_cache_dir
+
+
+# Generate a dataset by calling one of the supported external generators
+# TODO: Generator output cannot be used as dataset output directly, because of the
+# trailing columns.
+def generate_dataset(dataset):
```
joosthooz, line 27: Nice, now it should be possible to add type hints here without creating a circular dependency!
(or we could do that in #118 after merging this)

datalogistik/tpc_builders.py
```diff
@@ -173,10 +173,9 @@ def create_dataset(
                 # Not all tables will have enough rows for the full nr of partitions
                 if old_file.exists():
                     shutil.move(old_file, new_file)
-                    # reset permissions to read-only
-                    os.chmod(new_file, 0o444)
                     log.debug(f"Created {new_file}")
-            os.chmod(table_output_dir, 0o555)
+            # reset permissions to read-only
+            set_readonly_recurse(table_output_dir)
```
joosthooz, line 178: Nice, another circular dependency removed :+1:

joosthooz (approved):

datalogistik/generate_dataset.py
jonkeane: Let's do it in #118, if you don't mind

datalogistik/util.py
jonkeane: Done, there wasn't any section there about permissions (and this is really a pretty hidden feature regardless), but it is good to document things like this

PR #122, opened 2022-12-02
Lift the python pin

jonkeane:
Now that there are pyarrow wheels for python 3.11

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -22,6 +22,7 @@ on:
 
 env:
   NIGHTLY: https://pypi.fury.io/arrow-nightlies
+  PYTHON_VERSION: '3.11'
 
 jobs:
   check-copyright:
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -22,7 +22,6 @@ on:
 
 env:
   NIGHTLY: https://pypi.fury.io/arrow-nightlies
-  PYTHON_VERSION: '3.10'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/actions.yml
```diff
@@ -68,6 +68,7 @@ jobs:
         python-version:
           - '3.8'
           - '3.10'
+          - '3.11'
```
jonkeane, line 71: I'm not totally sure we need or want to keep 3.10 around like this, thoughts?

joosthooz: I don't have a strong opinion about it, we might want to use functionality from 3.11 like the `Self` typing feature but the 3.8 check would catch that kind of stuff.

joosthooz (approved):
Thanks!

.github/workflows/actions.yml
austin3dickey: Let's remove 3.10 from this list IMO

austin3dickey (approved):

.github/workflows/actions.yml
jonkeane: OK, I've removed it.

> we might want to use functionality from 3.11 like the Self typing feature but the 3.8 check would catch that kind of stuff.

We should be very careful about depending on too-new python versions since we want datalogistik to be easily installable in as many environments as possible. Honestly even our 3.8 dependency might be a little too high, but we can work to lower that if we run into it...

